PR #129, opened 2019-12-02
Fix #62

### LINKED ISSUE #62: static in array parameters
Thanks to Tachyon.

```c
$ cargo run
void f(int v[static 5]);
<stdin>:1:21: error: expected ')', got '5'
```

Should parse as an array known to have at least length 5.

pythondude325:
Fixes #62 and adds some unit tests.

### DIFF
--- a/src/parse/decl.rs
+++ b/src/parse/decl.rs
@@ -1834,18 +1834,18 @@ mod tests {
     fn test_pointers_and_arrays() {
         // cdecl: declare foo as array 10 of pointer to pointer to char
         assert!(match_type(
-            parse("char **foo[];"),
+            parse("char **foo[10];"),
             Array(
                 Box::new(Pointer(Box::new(Pointer(Box::new(Char(true)))))),
-                ArrayType::Unbounded,
+                ArrayType::Fixed(10),
             )
         ));
         // cdecl: declare foo as pointer to pointer to array 10 of int
         assert!(match_type(
-            parse("int (**foo)[];"),
+            parse("int (**foo)[10];"),
             Pointer(Box::new(Pointer(Box::new(Array(
                 Box::new(Int(true)),
-                ArrayType::Unbounded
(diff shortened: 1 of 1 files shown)


src/parse/decl.rs
```diff
@@ -992,6 +992,11 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
                             next: prefix.map(Box::new),
                         })
                     } else {
+                        if self.match_next(&Token::Keyword(Keyword::Static)).is_some() {
+                            // TODO: Add information for the compiler to know
+                            // about a possible optimization.
```
jyn514:
```suggestion
                            // to warn if `NULL` is passed into the function
```
This is the main benefit I see from `[static n]` declarations.

src/parse/decl.rs
```diff
@@ -1795,6 +1800,14 @@ mod tests {
                 ArrayType::Unbounded
             )
         ));
+        assert!(match_type(
+            parse("int a[5];"),
+            Array(Box::new(Int(true)), ArrayType::Fixed(5))
+        ));
+        assert!(match_type(
+            parse("char b[static 42];"),
+            Array(Box::new(Char(true)), ArrayType::Fixed(42))
```
jyn514: Hmm, I just realized this is not quite correct ... `static` should only be valid inside of a function prototype, like this: `int f(int a[static 5]);`. Anywhere else it should give a `semantic_err`.

The way you can tell you are in a function is when `allow_abstract` is set. This was originally intended for knowing whether `int f(int)` should be allowed, but can be repurposed here.

`allow_abstract` is not currently passed into `postfix_type`, you will have to pass it from `direct_declarator`.

jyn514: The error clang gives is `'static' used in array declarator outside of function prototype` which I think is fine to reuse.

PR #134, opened 2019-12-03
Static keyword in for array sizes in function parameters

### LINKED ISSUE #62: static in array parameters

pythondude325:
Allows the use of the `static` keyword for array sizes in function parameters.

Fixes #62

### DIFF
--- a/src/parse/decl.rs
+++ b/src/parse/decl.rs
@@ -1714,18 +1714,18 @@ mod tests {
     fn test_pointers_and_arrays() {
         // cdecl: declare foo as array 10 of pointer to pointer to char
         assert!(match_type(
-            parse("char **foo[];"),
+            parse("char **foo[10];"),
             Array(
                 Box::new(Pointer(Box::new(Pointer(Box::new(Char(true)))))),
-                ArrayType::Unbounded,
+                ArrayType::Fixed(10),
             )
         ));
         // cdecl: declare foo as pointer to pointer to array 10 of int
         assert!(match_type(
-            parse("int (**foo)[];"),
+            parse("int (**foo)[10];"),
             Pointer(Box::new(Pointer(Box::new(Array(
                 Box::new(Int(true)),
-                ArrayType::Unbounded
+                ArrayType::Fixed(10)
             )))),)
         ));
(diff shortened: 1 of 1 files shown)


src/parse/decl.rs
```diff
@@ -918,11 +918,12 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
     fn postfix_type(
         &mut self,
         mut prefix: Option<Declarator>,
+        allow_abstract: bool,
     ) -> Result<Option<Declarator>, SyntaxError> {
         // postfix
         while let Some(data) = self.peek_token() {
             prefix = match data {
-                // array
```
jyn514: lol this comment looks so sad. Thanks for changing it haha

src/parse/decl.rs
```diff
@@ -931,6 +932,20 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
                             next: prefix.map(Box::new),
                         })
                     } else {
+                        if let Some(keyword) = self.match_next(&Token::Keyword(Keyword::Static)) {
+                            if allow_abstract {
+                                // TODO: Add information for the compiler to know
+                                // to warn if `NULL` is passed into the function
+                            } else {
+                                return Err(SyntaxError(Locatable {
```
jyn514: This is a semantic error, not a syntactic one. The difference is that semantic errors are recoverable while syntax errors are not. Use `self.semantic_err` instead of returning an error from the function.

src/parse/decl.rs
```diff
@@ -1799,6 +1814,25 @@ mod tests {
...
+                    qualifiers: Default::default(),
+                    storage_class: Default::default(),
+                    init: true,
+                }],
+                varargs: false
+            })
+        ));
+
+        assert!(parse("int b[static 10];").unwrap().is_err());
+    }
```
jyn514: Looks great, thank you so much for adding this!

Issue #142, opened 2019-12-06
int() should decay to int(*)() in declarations


jyn514:
```c
int f(int g()) { return g(); }
int main() {
        return f(f);
}
<stdin>:3:19: error: invalid program: cannot implicitly convert 'int(*)(int() g)' to 'int()'
```

jyn514:
6.7.6.3, paragraph 8

jyn514:
Should probably go near https://github.com/jyn514/rcc/blob/master/src/parse/decl.rs#L863

PR #145, opened 2019-12-08
Inline function keyword

### LINKED ISSUE #84: Allow 'inline' keyword in declarations
For now, it should just be a no-op.

```c
static inline void f(); 
<stdin>:1:8: error: expected variable name or '(', got 'inline'
```

pythondude325:
Added the inline keyword and the proper error when used in inappropriate places.

Closes #84


src/data/mod.rs
```diff
@@ -173,18 +174,27 @@ impl Qualifiers {
...
     };
     pub const CONST_VOLATILE: Qualifiers = Qualifiers {
         c_const: true,
         volatile: true,
+        inline: false,
+    };
+    pub const INLINE: Qualifiers = Qualifiers {
+        c_const: false,
+        volatile: false,
+        inline: true,
```
jyn514: You don't need to add this unless you use it somewhere, the other `const` qualifiers are just helpers for unit tests mostly.

src/parse/decl.rs
```diff
@@ -139,12 +139,13 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
...
         };
-        if symbol.ctype.is_function() && qualifiers != Qualifiers::NONE {
+        if symbol.ctype.is_function() && (qualifiers.c_const || qualifiers.volatile) {
             warn(
                 &format!("{} has no effect on function return type", qualifiers),
                 id.location,
             );
-            qualifiers = Qualifiers::NONE;
+            qualifiers.c_const = false;
+            qualifiers.volatile = false;
```
jyn514: Good catch, I would have missed this!

src/parse/decl.rs
```diff
@@ -979,6 +981,20 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
...
+            ..
+        }) = prefix
+        {
+            // `inline` is allowed on function declarations
+        } else if qualifiers.inline {
+            self.semantic_err(
+                "`inline` is only allowed on function declarations",
+                self.last_location,
+            );
+        }
```
jyn514: So I know I keep changing my mind, but I think this would actually fit better in `Declarator::parse_type`. That way you don't have to keep passing the `qualifiers` down, you can just check them once at the end. Sorry for the inconvenience!

src/parse/decl.rs
```diff
@@ -209,7 +210,7 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
         self.expect(Token::Comma)?;
         loop {
             let decl = self
-                .declarator(false)?
+                .declarator(false, first_qualifiers)?
                 .expect("declarator should return Some when called with allow_abstract: false");
             let location = decl.id().unwrap().location;
             let (id, ctype) = decl
```
jyn514: This should also check that `inline` isn't set. From 6.7.4  Function specifiers, paragraph 2:

> Function specifiers shall be used only in the declaration of an identifier for a function.

src/parse/decl.rs
jyn514: Actually given all the other places you have to check maybe it would be better to just have each call site check it individually ...

src/parse/decl.rs
```diff
@@ -835,7 +836,7 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
             }
             let (sc, quals, param_type, _) = self.declaration_specifiers()?;
             // true: allow abstract_declarators
-            let declarator = self.declarator(true)?;
+            let declarator = self.declarator(true, quals)?;
```
jyn514: Technically `int f(inline int g());` should be an error but I don't think we need to catch it until https://github.com/jyn514/rcc/issues/142 is fixed.

src/parse/decl.rs
jyn514: I changed my mind again :laughing:  I think this is fine - it lets us check if we're in a function and the parameter decayed to a pointer. Specifically I'd like to give a better error than clang here:

```c
int f(inline int g());
```

clang: `<stdin>:1:7: error: 'inline' can only appear on functions`
gcc: `<stdin>:1:18: warning: parameter ‘g’ declared ‘inline’`
what I'd like to see: `error: 'inline' can only appear on functions. help: function parameters decay to pointers`

So long story short your current code is fine :)

src/parse/decl.rs
```diff
@@ -1895,7 +1895,7 @@ mod tests {
                 .unwrap()
                 .is_err()
         );
-        assert!(parse("typedef a inlint int;").unwrap().is_err());
+        assert!(parse("typedef a inline int;").unwrap().is_err());
```
jyn514: Maybe also check that this is a semantic error, not a syntax one?

Issue #131, opened 2019-12-02
[meta] Re-add auto-generated headers


jyn514:
https://github.com/jyn514/rcc/commit/6a03d0eb40430ec49c07ff4d65700358713d3693 is sad but necessary for now or CI tests will be useless. However, I would like to re-add autogenerated tests in the future.

jyn514:
Needs #127

PR #148, opened 2019-12-08
Parse bitfields; use system headers to test

jyn514:
It turns out bitfields were the only things missing!

Note that this only parses bitfields, they are not implemented at all (i.e. the code will be compiled exactly the same whether or not it uses bitfields).


jyn514:
Closes #131

PR #154, opened 2019-12-15
Retyped errors

pythondude325:
Changed all errors to use a new Compile Error trait capable of more things than the old system.

### DIFF
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -1,5 +1,6 @@
 use std::fmt;
 use std::any::Any;
+use std::fmt;
 use thiserror::Error;
 
 use super::{Expr, Locatable, Location};
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -82,16 +82,17 @@ impl<T, E> Recoverable for Result<T, (E, T)> {
     }
 }
 
-pub enum CompileErrorLevel {
+pub enum ErrorKind {
     Warning,
-    RecoverableError,
-    FatalError,
+    Lex,
+    Semantic,
+    Syntax,
 }
 
 pub trait NewCompileError: Any {
     // Idealy this should be an associated constant, but those are not allowed
     // in trait objects
-    fn level(&self) -> CompileErrorLevel;
+    fn kind(&self) -> ErrorKind;
 
     // Could also be an associated const
     fn name(&self) -> &'static str;
@@ -130,7 +131,7 @@ impl fmt::Display for (dyn NewCompileError) {
 }
 
 macro_rules! define_error {
-    ($error_short_name:expr, $error_level:expr, $error_name:ident($($arg_name:ident: $arg_type:ty ),*), $message:expr) => {
+    ($error_short_name:expr, $error_kind:expr, $error_name:ident($($arg_name:ident: $arg_type:ty ),*), $message:expr) => {
         #[derive(Debug)]
         pub struct $error_name {
             location: Location,
@@ -138,8 +139,8 @@ macro_rules! define_error {
         }
 
         impl $error_name {
-            pub fn new(location: Location, $($arg_name: $arg_type),*) -> $error_name {
-                $error_name { location, $($arg_name),* }
+            pub fn new(location: Location, $($arg_name: $arg_type),*) -> Box<dyn NewCompileError> {
+                Box::new($error_name { location, $($arg_name),* })
             }
         }
 
@@ -148,8 +149,8 @@ macro_rules! define_error {
                 $error_short_name
             }
 
-            fn level(&self) -> CompileErrorLevel {
-                $error_level
+            fn kind(&self) -> ErrorKind {
+                $error_kind
             }
 
             fn message(&self) -> String {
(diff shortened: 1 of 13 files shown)


src/data/error.rs
```diff
@@ -79,3 +80,136 @@ impl<T, E> Recoverable for Result<T, (E, T)> {
         }
     }
 }
+
+pub enum CompileErrorLevel {
+    Warning,
+    RecoverableError,
+    FatalError,
```
jyn514: I would rather these be called `Syntax` and `Semantic`, I think it better describes what they're for. Additionally, syntax errors aren't necessarily fatal.

Either we should get rid of `FatalError` or we should add a `Lex` variant.

src/data/error.rs
```diff
@@ -79,3 +80,136 @@ impl<T, E> Recoverable for Result<T, (E, T)> {
...
+}
+
+pub trait NewCompileError {
+    // Idealy this should be an associated constant, but those are not allowed
+    // in trait objects
+    fn level(&self) -> CompileErrorLevel;
```
jyn514: ~~This is not true: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3e1730a30bd1f73c3f7bc08e3ecdb84c~~ You meant `&dyn NewCompileError`, ignore me

src/data/error.rs
```diff
@@ -79,3 +80,136 @@ impl<T, E> Recoverable for Result<T, (E, T)> {
...
+
+    // Could also be an associated const
+    fn name(&self) -> &'static str;
+
+    fn message(&self) -> String;
+    fn location(&self) -> Location;
+}
+
+impl fmt::Debug for (dyn NewCompileError + 'static) {
```
jyn514: Why is this `'static`? That seems very limiting in case we want to reference a type or something.

src/data/error.rs
```diff
@@ -79,3 +80,136 @@ impl<T, E> Recoverable for Result<T, (E, T)> {
...
+}
+
+impl fmt::Display for (dyn NewCompileError + 'static) {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{} ({})", self.message(), self.name())
+    }
+}
+
+macro_rules! define_error {
+    ($error_short_name:expr, $error_level:expr, $error_name:ident($($arg_name:ident: $arg_type:ty ),*), $message:expr) => {
```
pythondude325: The usage here is:
```rust
define_error! {
    "short-name", error_level,
    ErrorName(arg_name: ArgType),
    "formatting string"
}
```

There's a little bit of magic in the type name, arguments, and formatting string.

src/data/error.rs
```diff
@@ -79,3 +80,136 @@ impl<T, E> Recoverable for Result<T, (E, T)> {
...
+    }
+
+    define_error! {
+        "syntax-error", CompileErrorLevel::FatalError,
+        GenericSyntaxError(message: String),
+        "invalid syntax: {message}"
+    }
+
+    define_error! {
+        "symantic-error", CompileErrorLevel::RecoverableError,
```
jyn514:
```suggestion
        "semantic-error", CompileErrorLevel::RecoverableError,
```

src/lex.rs
```diff
@@ -870,17 +871,12 @@ mod tests {
...
-    fn match_data_ref<T>(lexed: &Option<LexType>, closure: T) -> bool
-    where
-        T: FnOnce(Result<&Token, &str>) -> bool,
+        T: FnOnce(CompileResult<&Token>) -> bool,
     {
+        // TODO: This used to only accept lexing errors,
         match lexed {
-            Some(Ok(result)) => closure(Ok(&result.data)),
-            Some(Err(CompileError::Lex(result))) => closure(Err(&result.data)),
+            Some(Ok(l)) => closure(Ok(&l.data)),
```
jyn514:
```suggestion
            Some(Err(e)) if (e as &dyn Any).is::<errors::GenericLexError>() => closure(Err(e)),
```

src/data/error.rs
pythondude325: I think I did this because one of the errors said `Debug` (and others) were not implemented for `(dyn NewCompileError + 'static)` so I just implemented it with static without thinking about it.

src/data/error.rs
pythondude325: I always spell that wrong

src/lib.rs
```diff
@@ -37,7 +36,7 @@ mod parse;
 
 #[derive(Debug)]
 pub enum Error {
-    Source(VecDeque<CompileError>),
+    Source(VecDeque<Box<dyn NewCompileError>>),
```
jyn514: This double-boxes the errors. I would rather see `VecDeque<dyn NewCompileError>`.

src/data/error.rs
```diff
@@ -79,3 +81,174 @@ impl<T, E> Recoverable for Result<T, (E, T)> {
...
+
+impl PartialEq for (dyn NewCompileError) {
+    fn eq(&self, other: &Self) -> bool {
+        self.name() == other.name()
+    }
+}
+
+impl fmt::Display for (dyn NewCompileError) {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{} ({})", self.message(), self.name())
```
jyn514: I would rather this be
```suggestion
        write!(f, "{}: {}", self.name(), self.message())
```
On master, errors currently look like `:1:1: error: invalid program: cannot have empty program`. After your change, they say `:1:1: error: invalid program: cannot have empty program (semantic-error)`. Notice how it says both 'semantic-error' and 'invalid program' - it's telling you the same thing twice.

For the same reason, 'invalid program: ', etc. should be removed in the default errors below.

src/lib.rs
jyn514: This is not possible, `T` must implement `Sized`.

pythondude325:
Replaced with error enums in #162

Issue #150, opened 2019-12-10
Property testing


jyn514:
Mostly found from `rg expect src | rg -v 'self.expect' | rg -v expected`:

- `expr.rval()` should always be the same as `expr.rval().rval()`
- `t.is_assignment_operator()` implies `t.without_assignment()` is an `Ok` variant
- `Expr::Member` should always have `Struct` or `Union` type
- `Type::sizeof` and `Type::alignof` should always return the same `Result` variant
- `Expr::Compare` should only hold comparison tokens
- `last_saw_loop == false` <==> `!self.switches.is_empty()`
- `current_function` should be `Some` whenever parsing a statement
- `peek_next_token() == next_token()`

Various others I haven't bothered to write up.

https://github.com/AltSysrq/proptest (or maybe https://github.com/BurntSushi/quickcheck) seems like a good library to use.

jyn514:
Some of these can be solved completely with newtypes, like `Comparison` can very easily be an enum. Others sound to me like typestate (e.g. `current_function`) but I'd have to read up on typestate more since I haven't actually implemented it before. Others I really don't think can be part of the type system, like `peek_next_token`.

Issue #132, opened 2019-12-02
Clean up code


jyn514:
- Make struct_member_helper a function instead of a macro (or inline it)
- See if we can make TYPES a RwLock instead of a thread local RefCell
- probably lots more things

PR #155, opened 2019-12-15
Refactor

jyn514:
Addresses #150 , #132

### DIFF
--- a/src/data/lex.rs
+++ b/src/data/lex.rs
@@ -77,11 +77,9 @@ pub enum Keyword {
     Alignof,
 }
 
-#[derive(Clone, Debug, PartialEq)]
-pub enum Token {
-    PlusPlus,
-    MinusMinus,
-
+#[derive(Copy, Clone, Debug, PartialEq)]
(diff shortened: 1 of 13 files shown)


jyn514:
Main reasons for merging this:

- can get rid of `Expr::*_literal` functions
- can get rid of `is_assignment_operator`
- can remove `unwrap` and `expect` calls in https://github.com/jyn514/rcc/blob/8876bbde88ec432565525ec906438c0e425a370e/src/ir/expr.rs#L497, https://github.com/jyn514/rcc/blob/8876bbde88ec432565525ec906438c0e425a370e/src/ir/expr.rs#L503, https://github.com/jyn514/rcc/blob/8876bbde88ec432565525ec906438c0e425a370e/src/ir/expr.rs#L584, https://github.com/jyn514/rcc/blob/8876bbde88ec432565525ec906438c0e425a370e/src/parse/expr.rs#L120
- can add very nice `Expr::from` functions

Reasons not to merge this:
- it's so much code

PR #164, opened 2020-01-02
Error Handler

pythondude325:
This adds the `ErrorHandler` structure and makes some changes to error handling to improve code quality. I don't plan on merging until I have finished writing the docs for the error handler and related types.


src/parse/mod.rs
```diff
@@ -126,37 +129,43 @@ impl<I: Iterator<Item = Lexeme>> Iterator for Parser<I> {
...
-                    if self.pending.is_empty() {
-                        return Some(Err(err.into()));
+
+                // Check for end of file
+                if self.peek_token().is_none() {
+                    if self.scope.is_global() {
+                        self.leave_scope(self.last_location);
+                    }
+
+                    if self.pending.is_empty() && self.error_handler.is_empty() {
```
jyn514: The check to self.pending is redundant, you already checked it above. `self.leave_scope()` can only cause errors and there's no more tokens left in the stream anyway.

```suggestion
                    if self.error_handler.is_empty() {
```

src/parse/mod.rs
```diff
@@ -126,37 +129,43 @@ impl<I: Iterator<Item = Lexeme>> Iterator for Parser<I> {
...
+                        None
                     } else {
-                        self.pending.push_back(Err(err.into()));
-                        return self.pending.pop_front();
+                        self.next()
```
jyn514: Please avoid recursion unless necessary. The way you've written it now, `int i, j, k, ...` for 1000 elements or so at the end of the file will crash the parser.

Instead you can use `if let Some(err) = self.err_handler.pop_front() { ...` up above a little.

src/parse/mod.rs
```diff
@@ -126,37 +129,43 @@ impl<I: Iterator<Item = Lexeme>> Iterator for Parser<I> {
...
                     }
+                } else {
+                    match self.declaration() {
+                        Ok(decls) => {
+                            self.pending.extend(decls.into_iter());
+                        }
+                        Err(err) => {
+                            self.error_handler.push_err(err.into());
```
jyn514: nit: can you have `push_err` take `<T: Into<CompileError>>`? I imagine it will get rid of a lot of calls to `into()`.

jyn514: nit: can you have `push_err` take `<T: Into<CompileError>>`? I imagine it will get rid of a lot of calls to `into()`.

src/parse/mod.rs
```diff
@@ -415,9 +423,10 @@ mod tests {
         })
     }
     #[inline]
-    pub(crate) fn parser(input: &str) -> Parser<Lexer> {
-        let lex = Lexer::new("<test suite>".to_string(), input.chars(), false);
-        Parser::new(lex, false).unwrap()
+    pub(crate) fn parser(input: &str) -> Parser<impl Iterator<Item = Lexeme>> {
+        let tokens =
+            Lexer::new("<test suite>".to_string(), input.chars(), false).collect::<Vec<_>>();
+        Parser::new(tokens.into_iter(), false).unwrap()
```
jyn514: Why did you make this change? Now that they're using different error handlers there shouldn't be a problem with lifetimes.

src/parse/stmt.rs
```diff
@@ -441,7 +439,8 @@ mod tests {
     fn parse_stmt(stmt: &str) -> CompileResult<Option<Stmt>> {
         let mut p = parser(stmt);
         let exp = p.statement();
-        if let Some(Err(err)) = p.pending.pop_front() {
+        if !p.error_handler.is_empty() {
+            let err = p.error_handler.into_iter().next().unwrap();
```
jyn514: Not sure this has the same semantics but since it's test code I don't really care haha

src/data/error.rs
```diff
@@ -1,14 +1,59 @@
...
+}
+
+impl ErrorHandler {
+    /// Construct a new error handler.
+    pub(crate) fn new() -> ErrorHandler {
+        Default::default()
+    }
+
+    /// Add an error to the error handler.
+    pub(crate) fn push_err(&mut self, error: CompileError) {
```
jyn514: nit: can you change this to `push_back()`? We already know it's an error. See below for why `_back`.

src/data/error.rs
```diff
@@ -1,14 +1,59 @@
...
+        self.errors.push_back(error);
+    }
+
+    pub(crate) fn pop_err(&mut self) -> Option<CompileError> {
```
jyn514: Can you change this to `pop_front()`? I don't like that `handler.push(...); handler.pop(...)` modifies the state, it's not the same behavior as `Vec`.

src/data/error.rs
```diff
@@ -1,14 +1,59 @@
...
+        self.errors.pop_front()
+    }
+
+    /// Checks if there are errors in the handler
+    pub(crate) fn is_empty(&self) -> bool {
+        self.errors.is_empty()
+    }
+}
+
+impl IntoIterator for ErrorHandler {
```
jyn514: Is there a reason this has to be `IntoIterator` instead of `Iterator`? You could implement `next()` with `self.errors.pop_front()`.

I guess that would allow calling `push_err()` after you've already looped over it ... I don't know that that's a bad thing though.

src/parse/mod.rs
pythondude325: This is changed to collect the tokens into a Vec to match the way it happens in the main compile function. Should I change it back to using the lexer directly?

jyn514: Yes, I don't see a reason to change it.

pythondude325: I changed this because of how the main compile function collects into a `Vec` then passes the `Vec` to parser. This no longer has the lifetime issues that I originally made the change for, but the reasoning still stands. Should I change it back?

src/data/error.rs
pythondude325: This is only used for conveniently collecting into a Vec for some of the tests. So it doesn't matter to much.

src/parse/mod.rs
jyn514: Ok, now can you remove all the calls to `into()`? :P

src/parse/mod.rs
jyn514: Ok, now can you change all the others back? e.g. https://github.com/jyn514/rcc/pull/164/files#diff-b4aea3e418ccdb71239b96952d9cddb6R69

src/parse/decl.rs
```diff
@@ -2118,7 +2118,7 @@ mod tests {
     }
     #[test]
     fn bitfields() {
-        assert!(parse("struct { int:5; } a;").unwrap().is_err());
+        assert!(dbg!(parse("struct { int:5; } a;")).unwrap().is_err());
```
jyn514: Can you revert this? Or change it to `log::debug!`?

src/data/error.rs
pythondude325: Changed it to Iterator

src/data/mod.rs
```diff
@@ -581,18 +582,17 @@ mod tests {
...
-                    crate::Parser::new(
-                        crate::Lexer::new("<integration-test>", ty.chars(), false),
-                        false
-                    )
-                    .unwrap()
-                    .type_name()
-                    .unwrap()
-                    .data
-                    .0
+                    Parser::new(tokens.into_iter(), false)
```
jyn514: Could you change this back so it takes a lexer directly? Or at least remove the collect()? I know this is nitpicky.

src/parse/mod.rs
```diff
@@ -38,7 +38,7 @@ pub struct Parser<I: Iterator<Item = Lexeme>> {
     tokens: I,
     /// VecDeque supports pop_front with reasonable efficiency
     /// this is useful because errors are FIFO
```
jyn514:
```suggestion
     /// this is useful because there can be multiple declarators in a declaration: `int i, j, k;`
```

src/parse/mod.rs
```diff
@@ -126,37 +129,43 @@ impl<I: Iterator<Item = Lexeme>> Iterator for Parser<I> {
...
-                    if self.pending.is_empty() {
-                        return Some(Err(err.into()));
+
+                // Check for end of file
+                if self.peek_token().is_none() {
+                    if self.scope.is_global() {
+                        self.leave_scope(self.last_location);
+                    }
+
+                    if let Some(err) = self.error_handler.pop_front() {
```
jyn514: This would be better as `self.error_handler.pop_front().map(Err)`.

src/parse/mod.rs
```diff
@@ -126,37 +129,43 @@ impl<I: Iterator<Item = Lexeme>> Iterator for Parser<I> {
...
+                    match self.declaration() {
+                        Ok(decls) => {
+                            self.pending.extend(decls.into_iter());
+                        }
+                        Err(err) => {
+                            self.error_handler.push_back(err);
+                        }
+                    }
+
+                    self.next()
```
jyn514: I don't like this, I would rather use this:

```
                    self.declaration().map(|decls| {
                        self.pending.extend(decls.into_iter());
                        self.pending.pop_front()
                    })
```

src/parse/mod.rs
```diff
@@ -328,7 +324,18 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
         mem::replace(&mut self.next, tmp);
     }
     fn lex_error(&mut self, err: CompileError) {
-        self.pending.push_back(Err(err));
+        self.error_handler.push_back(err);
+    }
+    pub fn collect_results(self) -> (Vec<Locatable<Declaration>>, Vec<CompileError>) {
```
jyn514: This function can be removed in favor of [partition](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition). The call in main would look like `Parser::new(...).partition(|res| res.is_err())`.

jyn514: We decided against this because when we add warnings in the next PR we couldn't partition into 3 collections.

src/parse/mod.rs
jyn514: This doesn't work because for the behavior to be the same you have to rewrite `.next()`.

PR #172, opened 2020-01-11
Improve Parser::new() API

jyn514:
Pushes some of the burden onto the caller at the low low price of making
`Parser::new()` make much more sense

r? @pythondude325

### DIFF
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -78,6 +78,8 @@ pub enum Error {
 pub enum SemanticError {
     #[error("{0}")]
     Generic(String),
+    #[error("cannot have empty program")]
+    EmptyProgram,
 
     #[doc(hidden)]
     #[error("internal error: do not construct nonexhaustive variants")]
--- a/src/lex.rs
+++ b/src/lex.rs
@@ -125,6 +125,11 @@ impl<'a> Lexer<'a> {
             error_handler: ErrorHandler::new(),
         }
     }
+    /// Return the current location of the lexer
+    pub fn location(&self) -> Location {
+        self.location
+    }
+
     /// This lexer is somewhat unique - it reads a single character at a time,
     /// unlike most lexers which read a token at a time (e.g. string literals).
     /// This makes some things harder to do than normal, for example integer and float parsing, because
--- a/src/lex.rs
+++ b/src/lex.rs
@@ -1,6 +1,6 @@
 use lazy_static::lazy_static;
 
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 use std::convert::TryFrom;
 use std::str::Chars;
 
@@ -128,6 +128,21 @@ impl<'a> Lexer<'a> {
     /// Return the current location of the lexer
     pub fn location(&self) -> Location {
         self.location
+    }
+
+    /// Return the first valid token in the file,
+    /// or None if there are no valid tokens.
+    ///
+    /// In either case, return all invalid tokens found.
+    pub fn first_token(&mut self) -> (Option<Locatable<Token>>, VecDeque<CompileError>) {
+        let mut errs = VecDeque::new();
+        loop {
+            match self.next() {
+                Some(Ok(token)) => return (Some(token), errs),
+                Some(Err(err)) => errs.push_back(err),
+                None => return (None, errs),
+            }
+        }
     }
 
     /// This lexer is somewhat unique - it reads a single character at a time,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,23 +65,41 @@ pub fn compile(
     debug_ast: bool,
     debug_ir: bool,
 ) -> (Result<Product, Error>, VecDeque<CompileWarning>) {
-    let lexer = Lexer::new(filename, buf.chars(), debug_lex);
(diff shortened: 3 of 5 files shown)


src/parse/mod.rs
```diff
@@ -79,38 +79,23 @@ where
...
-            }
-        };
-        if !pending.is_empty() {
-            return Err(pending);
-        }
-        Ok(Parser {
-            scope: Scope::new(),
-            tag_scope: Scope::new(),
-            tokens: iter,
+    pub fn new(first: Locatable<Token>, tokens: I, debug: bool) -> Self {
```
pythondude325: Why do you have to pass the first token to `Parser::new`? Why can't it just call next on tokens itself?

src/data/error.rs
```diff
@@ -68,6 +68,8 @@ pub enum Error {
     // specific errors
     #[error("unterminated /* comment")]
     UnterminatedComment,
+    #[error("cannot have empty program")]
+    EmptyProgram,
```
pythondude325: The reign of specific errors begins

src/lex.rs
```diff
@@ -128,6 +128,11 @@ impl<'a> Lexer<'a> {
             error_handler: ErrorHandler::new(),
         }
     }
+    /// Return the current location of the lexer
+    pub fn location(&self) -> Location {
+        self.location
+    }
```
pythondude325: Is this just for ergonomics?

src/parse/mod.rs
pythondude325: I now realize that this is so that you can catch early lexer errors in the compile function. See comment in src/lib.rs

src/lib.rs
```diff
@@ -64,13 +64,40 @@ pub fn compile(
...
+                if lex_errs.is_empty() {
+                    lex_errs.push_back(CompileError::new(
+                        SourceError::EmptyProgram,
+                        lexer.location(),
+                    ));
+                }
+                return Err(Error::Source(lex_errs));
+            }
+        }
+    };
```
pythondude325: This situation seems pretty complicated for a public API. Maybe there should be a helper function?

src/lib.rs
```diff
@@ -102,3 +129,32 @@ pub fn link(obj_file: &Path, output: &Path) -> Result<(), io::Error> {
...
+                assert!(errs.is_empty());
+            }
+            _ => unreachable!(),
+        }
+        let err = compile("", "<test-suite>".into(), false, false, false)
+            .err()
+            .unwrap();
+        match err {
+            Error::Source(mut errs) => {
+                assert!(errs.pop_front().unwrap().data.kind() == ErrorKind::Semantic);
```
pythondude325: Can you also assert that this is an `EmptyProgram` error?

src/parse/mod.rs
```diff
@@ -79,38 +79,23 @@ where
     ///     If there is at least one token that is not an error, returns a parser.
     ///     Otherwise, returns a list of the errors.
     /// Otherwise, returns None.
```
pythondude325: This comment should be changed to reflect the current API

src/parse/mod.rs
```diff
@@ -326,7 +311,7 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
     fn lex_error(&mut self, err: CompileError) {
         self.error_handler.push_back(err);
     }
-    pub fn collect_results(self) -> (Vec<Locatable<Declaration>>, Vec<CompileError>) {
+    pub fn collect_results(&mut self) -> (Vec<Locatable<Declaration>>, Vec<CompileError>) {
```
pythondude325: Where do you need to use the parser after collecting the results from it?

src/lex.rs
jyn514: https://github.com/jyn514/rcc/pull/172/files#diff-b4aea3e418ccdb71239b96952d9cddb6R79

jyn514: The lexer location is private so people can't modify it. That would cause errors later when I implement spans because they could break the invariant that spans start before they end.

src/lib.rs
jyn514: Sure, I can add another `.collect_results()`.

src/parse/mod.rs
jyn514: https://github.com/jyn514/rcc/pull/172/files#diff-b4aea3e418ccdb71239b96952d9cddb6R92

pythondude325: If it's an empty program error isn't the location usually zero? Also for the empty program error I don't think it matters to the user where that error is, just the file

src/parse/mod.rs
jyn514: This is the whole point of the change. If it calls `next` on tokens itself, then `new()` has to return a `Result<_, Vec<CompileError>>`, there's no way to tell it 'no I'm sure this first token is valid'. See [`cpp`](https://github.com/jyn514/rcc/commit/80ee39f05efbb928852a5535bb7b8c2432ab2779#diff-a165b99f3b4afa48136cbfc5113ee601R171) for why this might be useful. Note also there that there's no way to return a `Vec` of errors, I'd have to add a new `error_handler` (although I'll probably have to end up doing that anyway).

Now that I think about it, maybe this isn't worth it ...

PR #178, opened 2020-01-13
Add spans to locations

### LINKED ISSUE #177: Use spans for error reporting
Partial start on this in https://github.com/jyn514/rcc/tree/spans. It needs to merge spans in stmts/expressions/declarations as well as fix various broken parts of the lexer (spans don't play well with `peek()` or `unput()`)

jyn514:
Closes #177.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,1 +1,3 @@
 edition = "2018"
+use_field_init_shorthand = true
+use_try_shorthand = true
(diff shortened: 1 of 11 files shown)


jyn514:
As much as I like `ranges`, it's probably not worth adding a dependency just for the one `.union()` function ... Let me take a look at the implementation and see if it would be hard to adapt.

PR #183, opened 2020-01-17
Location::with and Location::error

### LINKED ISSUE #181: Locatable::new should take Into<Error>
I have lots of code that looks like this:

```rust
Locatable::new(
    SemanticError::Generic("invalid preprocessing directive").into(),
    id.location,
),
```

It would be nice to write just `    SemanticError::Generic("invalid preprocessing directive")` instead.

pythondude325:
`Location::with` and `Location::error` functions to replace the `Locatable::new` function also used as `CompileError::new`

Fixes #181

### DIFF
--- a/src/parse/decl.rs
+++ b/src/parse/decl.rs
@@ -118,13 +118,10 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
                 )?))
             }
             (Type::Function(_), Some(t)) if *t == Token::EQUAL => {
-                return Err(Locatable::new(
-                    SyntaxError::from(format!(
-                        "expected '{{', got '=' while parsing function body for {}",
-                        symbol.id,
-                    )),
-                    id.location,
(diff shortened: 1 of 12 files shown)


src/parse/decl.rs
```diff
@@ -1123,18 +1119,20 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
...
-                    SyntaxError::from(format!("expected variable name or '(', got '{}'", x)),
-                    self.next_location(),
-                ));
+                // Construct and store the error so that the mutable borrow of
+                // self through x ends.
+                let error =
+                    SyntaxError::from(format!("expected variable name or '(', got '{}'", x));
+
+                // self can now be immutably borrowed to get the next location.
+                let err = Err(self.next_location().with(error));
```
jyn514: Honestly I think this was nicer before too :/

PR #184, opened 2020-01-17
First pass at preprocessor

jyn514:
Partially addresses #5


jyn514:
@pythondude325  Some of these are bug fixes that just happen to be required for implementing cpp. Let me know if you want those as separate PRs instead of mixed in with everything else.

jyn514:
Ok, most of the bugfixes should have been moved into master directly.

PR #189, opened 2020-01-18
fix crash dereferencing no-op

jyn514:
before, crashes looked like this:

```
Z[--*""];
thread 'main' panicked at 'not yet implemented: what's an lval but not a pointer or id? ""', src/parse/expr.rs:1123:18
```


jyn514:
Hmm, I'm not actually sure this is the right fix ... maybe noops should always be an rval?

jyn514:
Oh no this is right, Noops exist specifically so that we can change `lval` without changing anything else.

PR #206, opened 2020-01-19
Fix some panics in const folding

### LINKED ISSUE #38: Overflow in const fold
- [ ] addition (`0x7fffffffffffffffL + 1`, `-0x7fffffffffffffffL + -2`)
- [ ] subtraction (`-0x7fffffffffffffffL - 2`, `0x7fffffffffffffffL - -1`)
- [ ] multiplication (`0x7fffffffffffffffL * 2`, `(-0x7fffffffffffffffL - 1) * -1`)
- [ ] division (`1/0`, `(-0x7fffffffffffffffL - 1) / -1`)
- [ ] modulus (`1%0`, `(-0x7fffffffffffffffL - 1) % -1`)
- [x] shift left (`1 << 33`)
- [ ] negation(`-(-0x7fffffffffffffffL - 1L)`)

pythondude325:
This fixes some issues with integer overflows in constant folding and makes sure the correct behavior is used. Tests are also added.

Fixes #38

### DIFF
--- a/src/fold.rs
+++ b/src/fold.rs
@@ -1,6 +1,6 @@
 use crate::arch::CHAR_BIT;
 use crate::data::prelude::*;
-use std::ops::{Add, Sub, Div, Mul};
+use std::ops::{Add, Div, Mul, Sub};
 use Literal::*;
 
 macro_rules! fold_int_unary_op {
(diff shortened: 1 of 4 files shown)


src/fold.rs
```diff
@@ -24,17 +25,27 @@ macro_rules! fold_int_bin_op {
...
-            (_, _) => Ok(None),
+#[inline]
+fn fold_scalar_bin_op(
+    simple: fn(f64, f64) -> f64,
+    overflowing: fn(i64, i64) -> (i64, bool),
+    wrapping: fn(u64, u64) -> u64,
+) -> impl Fn(&Literal, &Literal, &Type) -> Result<Option<Literal>, Error> {
+    move |a: &Literal, b: &Literal, _ctype| match (a, b) {
+        (Int(a), Int(b)) => {
+            let (value, overflowed) = overflowing(*a, *b);
```
jyn514: Can you comment here that `overflowing` returns the wrapped value? I was confused the first time reading it.
```suggestion
            // `overflowing` returns the _wrapped_ value,
            //  so if we had negative overflow `value` would be positive and vice-versa.
            let (value, overflowed) = overflowing(*a, *b);
```

src/fold.rs
```diff
@@ -126,19 +126,32 @@ impl Expr {
...
             ExprType::Negate(expr) => expr.const_fold()?.map_literal(
+                &location,
                 |token| match token {
-                    Int(i) => Int(-i),
-                    UnsignedInt(u) => UnsignedInt(0u64.wrapping_sub(u)),
-                    Char(c) => Char(0u8.wrapping_sub(c)),
-                    Float(f) => Float(-f),
-                    _ => token,
+                    Int(i) => {
+                        let (value, overflowed) = i.overflowing_neg();
```
jyn514: If you think it's worth it, you could abstract this into a function and reuse it for `mod`. NBD if not though.

PR #222, opened 2020-01-21
Added specific errors to const-folding bit-shift operations

pythondude325:


src/fold.rs
```diff
@@ -500,39 +502,37 @@ fn shift_right(
...
-                semantic_err!(
-                    format!(
-                        "cannot shift left by {} or more bits for type '{}' (got {})",
-                        max_shift, ctype, shift
-                    ),
-                    *location,
-                );
+                return Err(location.error(SemanticError::TooManyShiftBits {
+                    is_left: true,
+                    current: shift,
```
jyn514: This doesn't print the type any more. I'm not tied to that, but is there a reason you removed it?

src/fold.rs
```diff
@@ -500,39 +502,37 @@ fn shift_right(
 fn shift_left(
     left: Expr,
     right: Expr,
-    ctype: &Type,
+    _ctype: &Type,
```
jyn514: If you're not going to use the type, please remove it from the function signature.

Issue #221, opened 2020-01-20
[ICE] crash on duplicate case labels


jyn514:
**Expected behavior**
RCC should give a semantic error.

**Actual Behavior**
RCC crashes during codegen.

**Code**

```c
int main() {
	switch(1) {
		case 1: case 1: return 1;
	}
}
thread 'main' panicked at 'Tried to set the same entry 1 twice', /home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/cranelift-frontend-0.56.0/src/switch.rs:58:9
```

<details><summary>Backtrace</summary>

```
  13: cranelift_frontend::switch::Switch::set_entry
  14: rcc::ir::stmt::<impl rcc::ir::Compiler>::case
  15: rcc::ir::stmt::<impl rcc::ir::Compiler>::compile_stmt
  16: rcc::ir::stmt::<impl rcc::ir::Compiler>::case
  17: rcc::ir::stmt::<impl rcc::ir::Compiler>::compile_stmt
  18: rcc::ir::stmt::<impl rcc::ir::Compiler>::compile_all
  19: rcc::ir::stmt::<impl rcc::ir::Compiler>::compile_stmt
  20: rcc::ir::stmt::<impl rcc::ir::Compiler>::switch
  21: rcc::ir::stmt::<impl rcc::ir::Compiler>::compile_stmt
  22: rcc::ir::stmt::<impl rcc::ir::Compiler>::compile_all
  23: rcc::ir::Compiler::compile_func
  24: rcc::ir::compile
  25: rcc::compile
  26: rcc::real_main
  27: rcc::main
```

</details>

Waiting on Cranelift to publish https://github.com/bytecodealliance/cranelift/pull/1358.

jyn514:
Hmm, I thought I fixed this in https://github.com/jyn514/rcc/tree/duplicate-case, but it still crashes on this input:

```c
int main() {
    switch(1)
      case 2: case 3: case 3:
        return 0;
}
thread 'main' panicked at 'Tried to set the same entry 3 twice', /home/joshua/.local/lib/cargo/git/checkouts/cranelift-e5a6a40e8f744a9d/765d3d8/cranelift-frontend/src/switch.rs:58:9
```

jyn514:
Whoop, I only put it in one arm of the else!

PR #223, opened 2020-01-21
Made all errors specific in stmt.rs

pythondude325:


src/ir/stmt.rs
```diff
@@ -357,16 +339,10 @@ impl Compiler {
...
-                        "'{}' statement not in loop or switch statement",
-                        if is_break { "break" } else { "continue" }
-                    ),
-                    location
-                );
+                Err(location.error(SemanticError::LoopExitOutsideLoopOrSwitch { is_break }))
             }
         } else if !is_break {
-            semantic_err!("'continue' not in loop".into(), location);
+            Err(location.error(SemanticError::ContinueNotInLoop))
```
pythondude325: I can combine these two errors if you want

src/data/error.rs
```diff
@@ -114,6 +114,27 @@ pub enum SemanticError {
...
+    LabelRedeclaration(cranelift::prelude::Ebb),
+
+    #[error("use of undeclared label {0}")]
+    UndeclaredLabel(crate::intern::InternedStr),
+
+    #[error("{}case outside of switch statement", if *(.is_default) { "default " } else { "" })]
+    CaseOutsideSwitch { is_default: bool },
+
+    #[error("cannot have multiple default cases in a switch statement")]
+    MultipleDefaultCase,
```
jyn514: This is part of a more general error which is "case repeated in switch statement". We don't catch that one currently but we will soon: https://github.com/jyn514/rcc/issues/221

pythondude325: We can change the error once the other duplicate cases are detected

jyn514: Can you change the name to DuplicateCase? It looks good after that.

jyn514: I'll do this myself when I fix https://github.com/jyn514/rcc/issues/221

PR #239, opened 2020-02-04
Fold cleanup

jyn514:
r? @pythondude325

### DIFF
--- a/src/fold.rs
+++ b/src/fold.rs
@@ -572,16 +572,19 @@ mod tests {
     fn test_const_fold(s: &str) -> CompileResult<Expr> {
         parse_expr(s).unwrap().const_fold()
     }
+    fn assert_fold(original: &str, expected: &str) {
+        assert_eq!(
+            test_const_fold(original).unwrap().expr,
+            test_const_fold(expected).unwrap().expr
+        )
+    }
 
     // I will be including the test cases from https://github.com/jyn514/rcc/issues/38#issue-491407941
     // as well as a working case for each operator
 
     #[test]
     fn test_addition() {
-        assert_eq!(
-            test_const_fold("3 + 4").unwrap().expr,
-            parse_expr("7").unwrap().expr
-        );
+        assert_fold("3 + 4", "7");
         assert_eq!(
             test_const_fold("0x7fffffffffffffffL + 1").unwrap_err().data,
             SemanticError::ConstOverflow { is_positive: true }.into()
@@ -596,10 +599,7 @@ mod tests {
 
     #[test]
     fn test_subtraction() {
-        assert_eq!(
-            test_const_fold("9 - 3").unwrap().expr,
-            parse_expr("6").unwrap().expr
-        );
+        assert_fold("9 - 3", "6");
         assert_eq!(
             test_const_fold("-0x7fffffffffffffffL - 2")
                 .unwrap_err()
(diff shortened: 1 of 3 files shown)


src/fold.rs
```diff
@@ -717,4 +741,19 @@ mod tests {
...
+        assert_eq!(
+            test_const_fold("'1' + '1'").unwrap().expr,
+            parse_expr("98").unwrap().expr,
+        );
+        assert_eq!(
+            test_const_fold("'1' % '1'").unwrap().expr,
+            parse_expr("0").unwrap().expr,
+        );
+        assert_eq!(
+            test_const_fold("'1' - 1.0").unwrap().expr,
```
pythondude325: It saddens me that this is valid C code.

src/fold.rs
```diff
@@ -418,21 +434,29 @@ fn cast(expr: Expr, ctype: &Type) -> CompileResult<ExprType> {
...
+
         (Float(f), Type::Bool) => Int((*f != 0.0) as i64),
         (Float(f), Type::Double) | (Float(f), Type::Float) => Float(*f),
         (Float(f), ty) if ty.is_integral() && ty.is_signed() => Int(*f as i64),
         (Float(f), ty) if ty.is_integral() => UnsignedInt(*f as u64),
+
+        (&Char(c), Type::Bool) => Int(c.into()),
+        (&Char(c), Type::Double) | (&Char(c), Type::Float) => Float(c.into()),
+        (&Char(c), ty) if ty.is_integral() && ty.is_signed() => Int(c.into()),
+        (&Char(c), ty) if ty.is_integral() => UnsignedInt(c.into()),
```
pythondude325: What is the reason that these are references and the others are not?

src/fold.rs
```diff
@@ -418,21 +434,29 @@ fn cast(expr: Expr, ctype: &Type) -> CompileResult<ExprType> {
 /// and we can use that to store the new type
 fn const_cast(token: &Literal, ctype: &Type) -> Option<Literal> {
```
pythondude325: Can you add a test for this function? Covering the old parts of it seem easy enough too.

src/fold.rs
jyn514: Otherwise `c` would be a `&u8` and I'd have to use `(*c).into()` which looks really ugly to my eye. I don't call `.into()` on the other variants so it's not a problem there.

src/fold.rs
jyn514: While writing test cases, I found that `!'0'` is incorrectly parsed as `'0' != 0` instead of `(int)'0' != 0`. I don't have time to fix it tonight, but I'll push the test cases I have in the meantime.

pythondude325: Are you going to do that in a separate PR?

jyn514: Got it! It wasn't a bug in `parse` at all, I wasn't folding casts to `char` properly.

jyn514:
This is ready for review.

src/fold.rs
```diff
@@ -548,16 +575,24 @@ mod tests {
     fn test_const_fold(s: &str) -> CompileResult<Expr> {
         parse_expr(s).unwrap().const_fold()
     }
+    fn assert_fold(original: &str, expected: &str) {
```
pythondude325: Good Idea

PR #248, opened 2020-02-07
Changed struct-related functions to take a StructType

pythondude325:
r? @jyn514

### DIFF
--- a/src/arch/mod.rs
+++ b/src/arch/mod.rs
@@ -31,7 +31,7 @@ lazy_static! {
 mod x64;
 pub use x64::*;
 
-/// Calculate the size of a union.
+/// Calculate the size of a union: the max of all member sizes
 pub fn union_size(struct_type: &StructType) -> Result<SIZE_T, &'static str> {
     let symbols = &struct_type.members();
     symbols
--- a/src/arch/mod.rs
+++ b/src/arch/mod.rs
@@ -13,7 +13,7 @@ use target_lexicon::Triple;
 
 use crate::data::{
     prelude::*,
-    types::{ArrayType, FunctionType},
+    types::{ArrayType, FunctionType, StructType},
 };
 use Type::*;
 
@@ -31,7 +31,9 @@ lazy_static! {
 mod x64;
 pub use x64::*;
 
-pub fn union_size(symbols: &[Symbol]) -> Result<SIZE_T, &'static str> {
+/// Calculate the size of a union.
+pub fn union_size(struct_type: &StructType) -> Result<SIZE_T, &'static str> {
+    let symbols = &struct_type.members();
     symbols
         .iter()
         .map(|symbol| symbol.ctype.sizeof())
@@ -39,13 +41,17 @@ pub fn union_size(symbols: &[Symbol]) -> Result<SIZE_T, &'static str> {
         .try_fold(1, |n, size| Ok(max(n, size?)))
 }
 
-pub fn struct_size(symbols: &[Symbol]) -> Result<SIZE_T, &'static str> {
+/// Calculate the size of a struct.
+pub fn struct_size(struct_type: &StructType) -> Result<SIZE_T, &'static str> {
+    let symbols = &struct_type.members();
     symbols.iter().try_fold(0, |offset, symbol| {
         Ok(Type::next_offset(offset, &symbol.ctype)?)
     })
 }
 
-pub fn struct_align(members: &[Symbol]) -> Result<SIZE_T, &'static str> {
+/// Calculate the alignment of a struct.
+pub fn struct_align(struct_type: &StructType) -> Result<SIZE_T, &'static str> {
+    let members = &struct_type.members();
     members.iter().try_fold(0, |max, member| {
         Ok(std::cmp::max(member.ctype.alignof()?, max))
     })
@@ -85,8 +91,8 @@ impl Type {
                     _ => return Err("enum cannot be represented in SIZE_T bits"),
                 })
             }
-            Union(struct_type) => union_size(&struct_type.members()),
-            Struct(struct_type) => struct_size(&struct_type.members()),
+            Union(struct_type) => union_size(&struct_type),
+            Struct(struct_type) => struct_size(&struct_type),
             Bitfield(_) => unimplemented!("sizeof(bitfield)"),
             // illegal operations
             Function(_) => Err("cannot take `sizeof` a function"),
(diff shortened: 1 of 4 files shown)


src/arch/mod.rs
```diff
@@ -121,9 +127,11 @@ impl Type {
     pub fn ptr_type() -> IrType {
         IrType::int(CHAR_BIT * PTR_SIZE).expect("pointer size should be valid")
     }
-    pub fn struct_offset(&self, members: &[Symbol], member: InternedStr) -> u64 {
+    /// Get the offset of the given struct member.
+    pub fn struct_offset(&self, struct_type: &StructType, member: InternedStr) -> u64 {
```
pythondude325: Do you know why this takes a struct type instead of just getting it from `self`? is it a remnant from earlier struct code?

jyn514: Good catch, this shouldn't take `self` at all; or rather, this should be in `impl StructType`, not `impl Type`.

src/arch/mod.rs
```diff
@@ -31,21 +31,27 @@ lazy_static! {
...
+    let symbols = &struct_type.members();
     symbols
         .iter()
         .map(|symbol| symbol.ctype.sizeof())
         // max of member sizes
         .try_fold(1, |n, size| Ok(max(n, size?)))
 }
 
-pub fn struct_size(symbols: &[Symbol]) -> Result<SIZE_T, &'static str> {
+/// Calculate the size of a struct.
```
jyn514:
```suggestion
/// Calculate the size of a struct: the sum of all member sizes 
```

src/arch/mod.rs
```diff
@@ -31,21 +31,27 @@ lazy_static! {
 mod x64;
 pub use x64::*;
 
-pub fn union_size(symbols: &[Symbol]) -> Result<SIZE_T, &'static str> {
+/// Calculate the size of a union.
```
jyn514:
```suggestion
/// Calculate the size of a union: the max of all member sizes
```

Also, this reminds me that I need to round this to the largest alignment. This is not necessarily a divisor of the largest size! Take for instance `union { long i; char a[3]; }`: the size is 6, but the alignment is 4.

src/arch/mod.rs
pythondude325: So Should I move this (and next_offset) to StructType?

jyn514: Yes, I think that would be better.

src/arch/mod.rs
```diff
@@ -31,24 +31,55 @@ lazy_static! {
...
+        for formal in members.iter() {
+            if formal.id == member {
+                return current_offset;
+            }
+            current_offset = Self::next_offset(current_offset, &formal.ctype)
+                .expect("structs should have valid size and alignment");
+        }
+        unreachable!("cannot call struct_offset for member not in struct");
+    }
+    /// Get the offset of the next truct member given the current offset.
```
jyn514:
```suggestion
    /// Get the offset of the next struct member given the current offset.
```

src/arch/mod.rs
```diff
@@ -31,24 +31,55 @@ lazy_static! {
...
-}
-
-pub fn struct_size(symbols: &[Symbol]) -> Result<SIZE_T, &'static str> {
-    symbols.iter().try_fold(0, |offset, symbol| {
-        Ok(Type::next_offset(offset, &symbol.ctype)?)
-    })
-}
+impl StructType {
+    /// Get the offset of the given struct member.
+    pub fn struct_offset(&self, member: super::InternedStr) -> u64 {
```
jyn514: Unrelated, but can you make this `pub(crate)`? Originally I just made everything `pub` but I've been trying to avoid that because a) it makes the API harder to use and b) the compiler doesn't warn you about unused code


```suggestion
    pub(crate) fn struct_offset(&self, member: super::InternedStr) -> u64 {
```

src/arch/mod.rs
```diff
@@ -31,24 +31,55 @@ lazy_static! {
...
+            if formal.id == member {
+                return current_offset;
+            }
+            current_offset = Self::next_offset(current_offset, &formal.ctype)
+                .expect("structs should have valid size and alignment");
+        }
+        unreachable!("cannot call struct_offset for member not in struct");
+    }
+    /// Get the offset of the next truct member given the current offset.
+    fn next_offset(mut current_offset: u64, ctype: &super::Type) -> Result<u64, &'static str> {
```
jyn514: Why did this change to `super::Type`?

src/arch/mod.rs
```diff
@@ -31,24 +31,55 @@ lazy_static! {
...
+    /// Calculate the size of a union: the max of all member sizes
+    pub fn union_size(&self) -> Result<SIZE_T, &'static str> {
+        let symbols = &self.members();
+        symbols
+            .iter()
+            .map(|symbol| symbol.ctype.sizeof())
+            // max of member sizes
+            .try_fold(1, |n, size| Ok(max(n, size?)))
+    }
+    /// Calculate the alignment of a struct.
```
jyn514:
```suggestion
    /// Calculate the alignment of a struct: the max of all member alignments
```

src/arch/mod.rs
```diff
@@ -110,7 +141,7 @@ impl Type {
             // Clang uses the largest alignment of any element as the alignment of the whole
             // Not sure why, but who am I to argue
             // Anyway, Faerie panics if the alignment isn't a power of two so it's probably for the best
-            Union(struct_type) | Struct(struct_type) => struct_align(&struct_type.members()),
+            Union(struct_type) | Struct(struct_type) => struct_type.struct_align(),
```
jyn514: You can probably rename this function to just `align()` now that it's an impl on `StructType`.

jyn514: Ditto for `struct_offset()` -> `offset()`.

src/arch/mod.rs
jyn514: Ditto for the other functions you changed.

src/arch/mod.rs
pythondude325: Its a remnant from `src/data/types.rs`

jyn514: Can you change it back? Ditto for `super::InternedStr`.

src/arch/mod.rs
```diff
@@ -33,7 +33,7 @@ pub use x64::*;
 
 impl StructType {
     /// Get the offset of the given struct member.
-    pub fn struct_offset(&self, member: super::InternedStr) -> u64 {
+    pub(crate) fn offset(&self, member: super::InternedStr) -> u64 {
```
jyn514:
```suggestion
    pub(crate) fn offset(&self, member: InternedStr) -> u64 {
```

PR #252, opened 2020-02-07
RecursionLimit error

pythondude325:
We don't need those `process::exit` shenanigans

### DIFF
--- a/src/parse/mod.rs
+++ b/src/parse/mod.rs
@@ -167,20 +167,20 @@ impl<I: Iterator<Item = Lexeme>> Iterator for Parser<I> {
     }
 }
 
+// this is just a guesstimate, it should probably be configurable
+#[cfg(debug_assertions)]
+pub(crate) const MAX_RECURSION_DEPTH: usize = 50;
+#[cfg(not(debug_assertions))]
+pub(crate) const MAX_RECURSION_DEPTH: usize = 200;
+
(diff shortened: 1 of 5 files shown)


src/parse/mod.rs
```diff
@@ -167,26 +167,22 @@ impl<I: Iterator<Item = Lexeme>> Iterator for Parser<I> {
...
         let guard = self.recursion_guard.clone();
         let depth = Rc::strong_count(&guard.0);
-        if depth > MAX_DEPTH {
-            eprintln!(
-                "fatal: maximum recursion depth exceeded ({} > {})",
-                depth, MAX_DEPTH
-            );
-            std::process::exit(102);
+        if depth > MAX_RECURSION_DEPTH {
+            return Err(self.last_location.with(SyntaxError::RecursionLimit(depth)));
```
jyn514:
```suggestion
            return Err(self.next_location().with(SyntaxError::RecursionLimit(depth)));
```

src/parse/mod.rs
```diff
@@ -167,26 +167,22 @@ impl<I: Iterator<Item = Lexeme>> Iterator for Parser<I> {
...
         }
-        guard
+        Ok(guard)
```
jyn514: There's no need for an early return here, you can use `if depth > max { Err() } else { Ok(guard) }`.

pythondude325:
I don't think this will be necessary after you rewrite the parser.

PR #258, opened 2020-02-08
Flesh out preprocessor

jyn514:
Still missing `#include`, it was harder than I thought. Leaving function macros for another day.

r? @pythondude325

### DIFF
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -4,7 +4,7 @@ use std::collections::{HashMap, VecDeque};
 use std::convert::TryFrom;
 
 use super::{Lexer, Token};
-use crate::data::error::CppError;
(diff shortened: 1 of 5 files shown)


src/lex/cpp.rs
```diff
@@ -279,9 +368,9 @@ impl<'a> PreProcessor<'a> {
     /// #if
```
pythondude325: Please improve this doc-comment

jyn514:
This has a bug (missing feature?): `#error can't do that` should give you `<stdin>:1:2: error: can't do that
` but instead gives an error because there's an apostrophe and it thinks it's the start of a character: `<stdin>:1:24 error: invalid token: Multi-character character literal`

PR #268, opened 2020-02-11
Add #include; fix bugs; add #elif; add defined()

jyn514:
This has everything except function macros, I think.


jyn514:
Needs a rebase.

PR #270, opened 2020-02-13
JIT Compilation support

playXE:
This PR adds JIT compilation support and simple API to JIT compile file and get access to functions and data of JITed module.


playXE:
I have question there: what we should do with `main` return value? Should we pass it to `std::process::exit`? Currently this value just ignored:
```rust
            let args = std::env::args();
            let argc = args.len() as i32;
            let vec_args = args
                .into_iter()
                .map(|string| std::ffi::CString::new(string).unwrap().as_ptr() as *const u8)
                .collect::<Vec<*const u8>>();
            let pointer = vec_args.as_ptr() as *const *const u8;
            let main: unsafe extern "C" fn(i32, *const *const u8) -> i32 =
                unsafe { std::mem::transmute(main_function) };
            let _ = unsafe { main(argc, pointer) };
```

src/main.rs
```diff
@@ -102,16 +102,37 @@ fn real_main(
...
+            let args = std::env::args();
+            let argc = args.len() as i32;
+            let vec_args = args
+                .into_iter()
+                .map(|string| std::ffi::CString::new(string).unwrap().as_ptr() as *const u8)
+                .collect::<Vec<*const u8>>();
+            let pointer = vec_args.as_ptr() as *const *const u8;
+            let main: unsafe extern "C" fn(i32, *const *const u8) -> i32 =
+                unsafe { std::mem::transmute(main_function) };
+            let _ = unsafe { main(argc, pointer) };
```
jyn514: Maybe it would be better to have a separate binary for JIT? I'm not sure I like the idea of RCC crashing at runtime if there's bad codegen.

jyn514: > I have question there: what we should do with main return value? Should we pass it to std::process::exit? Currently this value just ignored

Passing to exit() is better than before but I'm still not really happy with it. I would rather this be a library function somewhere that e.g. a REPL could call to get the return code.

jyn514:
Please add some tests as well - you can put them in a new directory `tests/runner-tests/jit`. I'll probably need to add a new mode to `tests/runner.rs` as well, I'll see if I have time on Friday.

Cargo.toml
```diff
@@ -51,3 +51,11 @@ debug-assertions = false
...
 incremental = false
 overflow-checks = false
+
+[[bin]]
+name = "rcc"
+path = "src/bin/main.rs"
+
+[[bin]]
+name = "rccjit"
+path = "src/bin/jit.rs"
```
jyn514: Newline at end of file please

src/bin/jit.rs
```diff
@@ -0,0 +1,397 @@
...
+            .map(|cstr| cstr.as_ptr() as *const u8)
+            .collect::<Vec<*const u8>>()
+            .as_ptr() as *const *const u8;
+        let main: unsafe extern "C" fn(i32, *const *const u8) -> i32 =
+            unsafe { std::mem::transmute(main_function) };
+        let _ = unsafe { main(argc, pointer) };
+    }
+    Ok(())
+}
+*/
```
jyn514: Remove commented out code please

benches/examples.rs
```diff
@@ -11,13 +11,18 @@ fn examples(c: &mut Criterion) {
...
     });
     group.finish();
-
     let mut group = c.benchmark_group("Factorial");
     group.bench_function("rcc", |b| {
-        b.iter(|| black_box(compile(FACTORIAL, &opts)));
+        b.iter(|| {
+            let module = initialize_aot_module("Factorial".to_owned());
+            black_box(compile(module, FACTORIAL, &opts))
+        });
```
jyn514: Good idea to make these benches jit, that's the latency that matters :)

src/bin/jit.rs
```diff
@@ -0,0 +1,397 @@
...
+        lex::Location,
+    },
+    preprocess, utils, Error, Opt, RccJIT,
+};
+use std::ffi::OsStr;
+
+static ERRORS: AtomicUsize = AtomicUsize::new(0);
+static WARNINGS: AtomicUsize = AtomicUsize::new(0);
+
+const HELP: &str = concat!(
```
jyn514: Hmm ... I know I told you to split the binaries up, but seeing all the duplicate code makes me unhappy... Maybe switch it back after all? I don't know what I want 😆


Certainly most of this can be reused, maybe you could make a bin_utils module or similar? I'm open to suggestions.

Cargo.toml
```diff
@@ -50,3 +51,8 @@ debug-assertions = false
 codegen-units = 1
 incremental = false
 overflow-checks = false
+
+[[bin]]
+name = "rcc"
+path = "src/bin/rcc.rs"
+
```
jyn514: Now there is an extra newline :P

src/bin/rcc.rs
```diff
@@ -102,16 +104,41 @@ fn real_main(
...
+        let (result, warnings) = compile(module, buf, &opt);
+        handle_warnings(warnings, file_id, file_db);
+        let mut rccjit = rcc::RccJIT::from_module(result?);
+        rccjit.finalize();
+        if let Some(main) = rccjit.get_compiled_function("main") {
+            let args = std::env::args();
+            let argc = args.len() as i32;
+            let vec_args = args
+                .map(|string| std::ffi::CString::new(string).unwrap())
+                .collect::<Vec<std::ffi::CString>>();
```
jyn514:
```suggestion
                .collect::<Vec<_>>();
```

src/bin/rcc.rs
```diff
@@ -102,16 +104,41 @@ fn real_main(
...
+            let pointer = vec_args
+                .iter()
+                .map(|cstr| cstr.as_ptr() as *const u8)
+                .collect::<Vec<*const u8>>()
```
jyn514:
```suggestion
                .collect::<Vec<_>>()
```

src/bin/rcc.rs
```diff
@@ -102,16 +104,41 @@ fn real_main(
...
+        assemble(product, tmp_file.as_ref())?;
+        link(tmp_file.as_ref(), output).map_err(io::Error::into)
+    } else {
+        let module = rcc::initialize_jit_module();
+        let (result, warnings) = compile(module, buf, &opt);
+        handle_warnings(warnings, file_id, file_db);
+        let mut rccjit = rcc::RccJIT::from_module(result?);
+        rccjit.finalize();
+        if let Some(main) = rccjit.get_compiled_function("main") {
+            let args = std::env::args();
```
jyn514: I think this will include `rcc` as the first argument, could you use `args()[1..]` instead?

src/bin/rcc.rs
```diff
@@ -102,16 +104,41 @@ fn real_main(
...
+            let vec_args = args
+                .map(|string| std::ffi::CString::new(string).unwrap())
+                .collect::<Vec<std::ffi::CString>>();
+            let pointer = vec_args
+                .iter()
+                .map(|cstr| cstr.as_ptr() as *const u8)
+                .collect::<Vec<*const u8>>()
+                .as_ptr() as *const *const u8;
+            let main: unsafe extern "C" fn(i32, *const *const u8) -> i32 =
+                unsafe { std::mem::transmute(main) };
```
jyn514: What is this transmute doing? Why is it safe?

src/bin/rcc.rs
```diff
@@ -102,16 +104,41 @@ fn real_main(
...
+            let _ = unsafe { main(argc, pointer) };
```
jyn514: As discussed yesterday, please call `std::process::exit` with this return code.

src/ir/mod.rs
```diff
@@ -97,44 +87,26 @@ pub(crate) fn compile(
...
-            cranelift_module::default_libcall_names(),
-        )
-        .expect("unknown error creating module");
+let builder = ObjectBuilder::new(
+    isa,
+    name,
+    ObjectTrapCollection::Disabled,
+    cranelift_module::default_libcall_names(),
+)
+.expect("unknown error creating module");*/
```
jyn514: Please remove this commented out code.

src/lib.rs
```diff
@@ -127,8 +130,68 @@ pub fn preprocess(
...
+    let isa = cranelift::codegen::isa::lookup(arch::TARGET)
+        .unwrap_or_else(|_| utils::fatal(format!("platform not supported: {}", arch::TARGET), 5))
+        .finish(cranelift::codegen::settings::Flags::new(flags_builder));
+
+    let builder = SimpleJITBuilder::with_isa(isa, cranelift_module::default_libcall_names());
+    let module = Module::new(builder);
+    module
+}
+
+pub fn initialize_aot_module(name: String) -> Module<ObjectBackend> {
```
jyn514: These two functions are mostly the same, can you refactor them to have less duplicated code?

src/lib.rs
```diff
@@ -219,6 +282,55 @@ pub fn link(obj_file: &Path, output: &Path) -> Result<(), io::Error> {
...
+
+    /// Invoke this function before trying to get access to "new" compiled functions.
+    pub fn finalize(&mut self) {
+        self.module.finalize_definitions();
+    }
+    /// Get compiled function, if this function doesn't exist then `None` is returned, otherwise it's address returned.
+    ///
+    /// # Panics
+    /// Panics if function is not compiled (finalized). Try to invoke `finalize` before using `get_compiled_function`.
+    pub fn get_compiled_function(&mut self, name: &str) -> Option<*const u8> {
```
jyn514: Is there a reason you didn't call `finalize()` here instead of having the user do it?

playXE: > Is there a reason you didn't call `finalize()` here instead of having the user do it?

There are no function in `Module` that may say is module finalized or no, and users may not need to finalize all functions in module.

src/lib.rs
playXE: Yes,I will do this.

src/bin/rcc.rs
```diff
@@ -125,17 +125,22 @@ fn real_main(
...
+                let pointer = vec_args
+                    .iter()
+                    .map(|cstr| cstr.as_ptr() as *const u8)
+                    .collect::<Vec<*const u8>>()
+                    .as_ptr() as *const *const u8;
+                let main: unsafe extern "C" fn(i32, *const *const u8) -> i32 =
+                    unsafe { std::mem::transmute(main) }; // this transmute is safe: this function is finalized(`rccjit.finalize()`) and **guaranteed** to be non-null
+                unsafe { main(argc, pointer) }
+            };
+            std::process::exit(exit_code);
```
jyn514: Now that you've mentioned releasing resources, this doesn't release `Files` either ... But we can fix that in a follow-up PR, it should be done by the OS anyway.

src/lib.rs
```diff
@@ -317,13 +327,43 @@ impl RccJIT {
...
+        // CString should be alive if we want to pass it's pointer to another function properly, otherwise this may lead to memory leak or UB.
+        let pointer = vec_args
+            .iter()
+            .map(|cstr| cstr.as_ptr() as *const u8)
+            .collect::<Vec<_>>()
+            .as_ptr() as *const *const u8;
+        assert_ne!(main, std::ptr::null());
+        // this transmute is safe: this function is finalized(`self.finalize()`) and **guaranteed** to be non-null
+        let main: unsafe extern "C" fn(i32, *const *const u8) -> i32 = std::mem::transmute(main);
+        // through transmute is safe,invoking this function is unsafe because we invoke C code.
```
jyn514:
```suggestion
        // though transmute is safe, invoking this function is unsafe because we invoke C code.
```

src/lib.rs
```diff
@@ -317,13 +327,43 @@ impl RccJIT {
...
+    /// It can segfault, access out-of-bounds memory, cause data races, or do anything else C can do.
+    pub unsafe fn run_main(&mut self) -> Option<i32> {
+        self.finalize();
+        let main = self.get_compiled_function("main")?;
+        let args = std::env::args().skip(1);
+        let argc = args.len() as i32;
+        let vec_args = args
+            .map(|string| std::ffi::CString::new(string).unwrap())
+            .collect::<Vec<_>>();
+        // CString should be alive if we want to pass it's pointer to another function properly, otherwise this may lead to memory leak or UB.
```
jyn514:
```suggestion
        // CString should be alive if we want to pass its pointer to another function properly, otherwise this may lead to memory leak or UB.
```

jyn514:
Closing in favor of #322

PR #273, opened 2020-02-13
add _Noreturn

jyn514:
This makes tests stop failing locally on my machine.

### DIFF
--- a/src/data/mod.rs
+++ b/src/data/mod.rs
@@ -162,23 +162,41 @@ pub struct Symbol {
 pub struct Qualifiers {
     pub volatile: bool,
     pub c_const: bool,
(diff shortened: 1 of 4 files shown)


src/data/mod.rs
```diff
@@ -162,25 +162,43 @@ pub struct Symbol {
 pub struct Qualifiers {
     pub volatile: bool,
     pub c_const: bool,
+    pub func: FunctionQualifiers,
+}
+
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Hash)]
+pub struct FunctionQualifiers {
```
pythondude325: Good idea moving these two into a separate struct.

PR #274, opened 2020-02-13
update hfuzz binary

Byter09:


fuzz/fuzz_targets/hfuzz.rs
```diff
@@ -1,17 +1,20 @@
...
+use rcc::{Files, Locatable};
 use rcc::data::lex::{Keyword::*, Token};
-use rcc::Locatable;
 use rcc::PreProcessor;
+use std::default::Default;
 
-fn lex(s: &str) -> PreProcessor {
-    PreProcessor::new("<test-suite>", s.chars(), false)
+fn lex<'s>(s: &'s str, file: FileId, files: &'s mut Files) -> PreProcessor<'s> {
+    PreProcessor::new(file, s, false, files)
```
jyn514: Honestly at this point it probably makes sense to get rid of `lex()` altogether, we're writing way more code then we're saving

Byter09: done

PR #275, opened 2020-02-14
Doc comments for src/arch/mod.rs

pythondude325:
I realized you made this module private, but I already wrote the comments, so I will push them.

### DIFF
--- a/src/arch/mod.rs
+++ b/src/arch/mod.rs
@@ -17,14 +17,18 @@ use crate::data::{
 };
 use Type::*;
 
-// NOTE: this is required by the standard to always be one
+/// Size of a `char` in bytes;
+/// 
+/// Note: The standard requires this to always be one.
 const CHAR_SIZE: u16 = 1;
 
 // TODO: allow this to be configured at runtime
+/// The target triple for the configured target.
 pub(crate) const TARGET: Triple = Triple::host();
 // TODO: make this const when const_if_match is stabilized
 // TODO: see https://github.com/rust-lang/rust/issues/49146
 lazy_static! {
+    /// The calling convention for the current target.
     pub(crate) static ref CALLING_CONVENTION: CallConv = CallConv::triple_default(&TARGET);
 }
 
--- a/src/arch/mod.rs
+++ b/src/arch/mod.rs
@@ -19,13 +19,19 @@ use crate::data::{
 };
 use Type::*;
 
-/// Size of a `char` in bytes;
+/// Size of a `char` in bytes
 ///
 /// Note: The standard requires this to always be one.
+/// http://port70.net/~nsz/c/c11/n1570.html#6.5.3.5
 const CHAR_SIZE: u16 = 1;
 
 // TODO: allow this to be configured at runtime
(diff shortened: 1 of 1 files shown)


src/arch/mod.rs
```diff
@@ -17,14 +19,18 @@ use crate::data::{
 };
 use Type::*;
 
-// NOTE: this is required by the standard to always be one
+/// Size of a `char` in bytes;
```
jyn514:
```suggestion
/// Size of a `char` in bytes
```

src/arch/mod.rs
```diff
@@ -17,14 +19,18 @@ use crate::data::{
...
+///
+/// Note: The standard requires this to always be one.
```
jyn514:
```suggestion
/// Note: The standard requires this to always be one.
/// http://port70.net/~nsz/c/c11/n1570.html#6.5.3.4
```

src/arch/mod.rs
```diff
@@ -17,14 +19,18 @@ use crate::data::{
...
 const CHAR_SIZE: u16 = 1;
 
 // TODO: allow this to be configured at runtime
+/// The target triple for the configured target.
```
jyn514: I don't think this is very helpful. Maybe link to an explanation of target triple instead: https://clang.llvm.org/docs/CrossCompilation.html#target-triple or https://docs.rs/target-lexicon/0.10.0/target_lexicon/struct.Triple.html

Also, this is the host target, not any configured target.

src/arch/mod.rs
```diff
@@ -83,6 +89,7 @@ impl StructType {
 }
 
 impl Type {
+    /// Returns true if `other` can be cast to `self` without losing infomation.
```
jyn514:
```suggestion
    /// Returns true if `other` can be converted to `self` without losing infomation.
```

Cast makes me think an explicit cast (`(int)1.4`), I think conversion is better.

src/arch/mod.rs
```diff
@@ -149,9 +160,11 @@ impl Type {
             Error => Err("cannot take `alignof` <type error>"),
         }
     }
+    /// Get the pointer type
```
jyn514: This is not really an accurate description. This returns an integer type which is large enough to represent the pointer; this is an IR-level type and has no relation to any C type.

src/arch/mod.rs
```diff
@@ -149,9 +160,11 @@ impl Type {
...
     pub fn ptr_type() -> IrType {
         IrType::int(CHAR_BIT * PTR_SIZE).expect("pointer size should be valid")
     }
+    /// Convert `self` into the type for the IR.
```
jyn514:
```suggestion
    /// Return an IR type which can represent this C type
```

'Convert' suggests that it consumes `self`.

src/arch/mod.rs
pythondude325: My new explanation for target triple may be a little long winded.

jyn514: I like it :)

Issue #122, opened 2019-11-28
Support hex character literals


jyn514:
e.g. `char c = '\x80'`. This requires `parse_single_char` to return a `u8`, not a `char`, since you can have values between 128 and 255.

http://port70.net/~nsz/c/c11/n1570.html#6.4.4.4

jyn514:
We may as well support octal literals at the same time (`'\10'` is `'\x08'`).

jyn514:
Proposed algorithm:

1. On seeing a `\<digit>`, call `parse_int` with a `radix` of 8 and a `start` of the digit seen.
2. On seeing a `\x<digit>`, call `parse_int` with a radix of 16.

Shouldn't be too hard actually.

jyn514:
Hold on, that doesn't work because these escapes have a fixed number of digits.

```
<stdin>:1:10: warning: multi-character character constant [-Wmultichar]
char z = '\0005';
         ^
```

Hmm ... maybe take digits one at a time up until a max? I wonder if that could be added into `parse_int` (and whether I want to). Logos is looking awfully good ...

Note that the max for octals is 3 digits (`\377`) and the max for hex is 2 (`\xff`).

jyn514:
Apparently hex digits don't have a max ??

>  
```
         octal-escape-sequence:
                 \ octal-digit
                 \ octal-digit octal-digit
                 \ octal-digit octal-digit octal-digit
          hexadecimal-escape-sequence:
                 \x hexadecimal-digit
                 hexadecimal-escape-sequence hexadecimal-digit
```

PR #283, opened 2020-02-19
Add hex and octal escapes

jyn514:
Closes https://github.com/jyn514/rcc/issues/122

r? @pythondude325

### DIFF
--- a/src/lex/mod.rs
+++ b/src/lex/mod.rs
@@ -493,8 +493,10 @@ impl Lexer {
         let mut base = 0_u64;
         let mut update = |this: &mut Self, c: u8| {
             this.next_char();
-            // TODO: handle overflow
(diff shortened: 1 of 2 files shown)


src/lex/mod.rs
```diff
@@ -424,19 +426,32 @@ impl Lexer {
                     Ok(match c {
                         // escaped newline: "a\
                         // b"
-                        b'\n' => return self.parse_single_char(string),
+                        b'\n' => unreachable!("should be handled earlier"),
```
jyn514: This is a holdover from when I changed backslashes to be handled earlier: https://github.com/jyn514/rcc/blob/519345079a25f6b6f9b2b267893243c29cb1788e/src/lex/mod.rs#L86

PR #300, opened 2020-02-26
Warn on other declarations that default to integer type

### LINKED ISSUE #225: Missing warning for `*p;`
```c
i;
*p;
<stdin>:2:1: warning: type specifier missing, defaults to int
i;
^
```

repnop:
Fixes #225

### DIFF
--- a/src/parse/decl.rs
+++ b/src/parse/decl.rs
@@ -443,8 +443,8 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
             Some(ctype) => ctype,
             None => {
                 if signed.is_none() {
-                    // these are the only tokens that can come before an id in a
-                    // declarator if it's not a declarator, it's invalid anyway
+                    // these are the only tokens that can come before an id in a declarator
+                    // if it's not a declarator, it's invalid anyway
                     // other parts of the parser will have a better error
                     // message
                     match self.peek_token() {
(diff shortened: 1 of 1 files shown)


src/parse/decl.rs
```diff
@@ -445,10 +445,13 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
                 if signed.is_none() {
                     // if it's not an id, it's invalid anyway
```
jyn514:
```suggestion
                    // these are the only tokens that can come before an id in a declarator
                    // if it's not a declarator, it's invalid anyway
```

repnop: Changed :)

src/parse/decl.rs
```diff
@@ -443,8 +443,10 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
             Some(ctype) => ctype,
             None => {
                 if signed.is_none() {
-                    // if it's not an id, it's invalid anyway
-                    // other parts of the parser will have a better error message
+                    // these are the only tokens that can come before an id in a
+                    // declarator if it's not a declarator, it's invalid anyway
```
jyn514: You line wrapped it and it changed the meaning :(
```suggestion
                    // these are the only tokens that can come before an id in a declarator
                    // if it's not a declarator, it's invalid anyway
```

repnop: oh no! hecc

PR #302, opened 2020-02-27
Print out parsing errors if we're going to crash

jyn514:
Note that this doesn't do anything about lexing errors because the parser doesn't know the type of the preprocessor.


src/parse/mod.rs
```diff
@@ -184,6 +184,16 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
                 "fatal: maximum recursion depth exceeded ({} > {})",
                 depth, MAX_DEPTH
             );
+            if !self.error_handler.is_empty() {
+                println!("pending errors:");
+                // we're about to crash, no point worrying about performance
+                for error in self.error_handler.clone() {
```
pythondude325: Why not simple avoid this clone?

jyn514: `recursion_check` takes `&self`, not `&mut self`.

pythondude325: But why not clone to a temp variable and use that error handler twice?

src/parse/mod.rs
```diff
@@ -184,6 +184,16 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
...
+                    println!("- error: {}", error.data);
+                }
+                for warning in self.error_handler.clone().warnings {
+                    println!("- warning: {}", warning.data);
+                }
```
pythondude325:
```suggestion
                // we're about to crash, no point worrying about performance
                let error_handler = self.error_handler.clone();
                for error in error_handler {
                    println!("- error: {}", error.data);
                }
                for warning in error_handler.warnings {
                    println!("- warning: {}", warning.data);
                }
```

PR #306, opened 2020-02-28
Return LexErrors from `Lexer::next()` instead of CompileErrors

jyn514:


src/data/error.rs
```diff
@@ -490,11 +510,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn test_compile_error_from_locatable_string() {
-        let _ = CompileError::from(Location::default().with("apples".to_string()));
```
pythondude325: What was the reason to remove this test?

src/lex/cpp.rs
```diff
@@ -210,7 +210,11 @@ impl<'a> PreProcessor<'a> {
     }
     #[inline]
     fn next_token(&mut self) -> Option<CppResult<Token>> {
-        self.lexer_mut().next()
+        match self.lexer_mut().next() {
+            Some(Err(err)) => Some(Err(err.into())),
+            Some(Ok(ok)) => Some(Ok(ok)),
+            None => None,
```
pythondude325: This could be replaced be some `map`s, do you prefer the `match` for some reason?

src/lex/mod.rs
```diff
@@ -654,7 +656,7 @@ impl Lexer {
 impl Iterator for Lexer {
     // option: whether the stream is exhausted
     // result: whether the next lexeme is an error
-    type Item = CompileResult<Locatable<Token>>;
+    type Item = Result<Locatable<Token>, Locatable<LexError>>;
```
pythondude325: Use the `LexResult` alias here.

jyn514:
Addressed review comments.

PR #307, opened 2020-02-28
Allow setting macros at the command line

jyn514:
~~Waiting on #305~~ merged. ~~The error type for `impl TryFrom<&str> for Vec<Token>` should be changed to LexError when #306 is merged~~ done. ~~Needs to remove the `unwrap()` in main (```-D 'a=`'``` will cause a crash)~~ done.


jyn514:
Ok, so it looks like gcc and clang allow `-D` to be an arbitrary string that's only lexed if the macro appears in the program? I don't know if I like that, it makes it really hard to fit into the existing code.
```
$ run_gcc -D'a=`'
a
<command-line>:0:3: error: stray ‘`’ in program
<stdin>:1:1: note: in expansion of macro ‘a’
```

PR #308, opened 2020-02-28
update codespan in fuzzing binaries

Byter09:


jyn514:
Hopefully https://github.com/jyn514/rcc/pull/305 will fix this more permanently. It wouldn't have helped in this case anyway though ... Maybe rcc should re-export codespan? Do you mind changing this PR to do that instead?

Byter09:
As talked about in discord, closing this in favour of a re-export PR.

PR #317, opened 2020-03-02
Add script to report issues

jyn514:
r? @Byter09

You don't need to read the code, I just want to make sure it behaves roughly the way you expect (documented in CONTRIBUTING.md).


jyn514:
Right now it will prompt you if there's no browser available, I can change that to just print the URL instead. I might also add an option to never open the browser and just print it out.

jyn514:
Also considering inlining the python script so you can copy this anywhere and use it for other projects.

PR #318, opened 2020-03-02
parser proptests

Byter09:
This is part of #150.

### DIFF
--- a/src/data/lex.rs
+++ b/src/data/lex.rs
@@ -2,11 +2,13 @@ use std::cmp::Ordering;
 
 use codespan::{FileId, Span};
 use cranelift::codegen::ir::condcodes::{FloatCC, IntCC};
-#[cfg(test)] use proptest_derive::Arbitrary;
+#[cfg(test)]
+use proptest_derive::Arbitrary;
+
+#[cfg(test)]
+use test::arb_interned_str;
 
 use crate::intern::InternedStr;
-
-#[cfg(test)] use self::test::arb_interned_str;
 
 // holds where a piece of code came from
 // should almost always be immutable
--- a/src/data/lex.rs
+++ b/src/data/lex.rs
@@ -1,8 +1,12 @@
-use crate::intern::InternedStr;
+use std::cmp::Ordering;
 
 use codespan::{FileId, Span};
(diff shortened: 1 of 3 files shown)


src/data/lex.rs
```diff
@@ -222,7 +232,6 @@ impl Literal {
     }
 }
 
-use cranelift::codegen::ir::condcodes::{FloatCC, IntCC};
```
jyn514: Good call, thanks.

src/parse/mod.rs
```diff
@@ -10,6 +6,10 @@ use std::rc::Rc;
 
 use crate::data::{prelude::*, Scope};
 
+mod decl;
+mod expr;
+mod stmt;
+
```
jyn514: Why did you move these?

src/parse/mod.rs
```diff
@@ -488,4 +490,39 @@ pub(crate) mod tests {
         let buf: String = buf.into_iter().collect();
         assert!(parse(&buf).is_none());
     }
+
+    prop_compose! {
+        fn arb_vec_result_locatable_token()(tokens in any::<Vec<Token>>()) -> Vec<CompileResult<Locatable<Token>>> {
+            tokens.into_iter().map(|token| Ok(Locatable { data: token, location: Location::default()})).collect()
+        }
```
jyn514: I don't think this should be its own `prop!` implementation, it never generates any error variants. Could you make it a free function instead?

src/parse/mod.rs
```diff
@@ -488,4 +490,39 @@ pub(crate) mod tests {
...
+    }
+
+    proptest! {
+        #[test]
+        fn proptest_peek_equals_token(
+            first in any::<Token>(),
+            tokens in arb_vec_result_locatable_token()
+            ) {
+            let mut parser = Parser::new(Locatable { data: first, location: Location::default() }, tokens.into_iter(), false);
```
jyn514: It would be nice for this to be a `prop!` impl eventually but we don't need it for a first pass.

src/parse/mod.rs
jyn514: Actually I changed my mind - we can just add error variants in the future.

PR #319, opened 2020-03-03
Replaced hexf with hexponent

pythondude325:
This first revision is a 1:1 replacement. Improvements can be made by using features that `hexponenet` has and `hexf` doesn't.

### DIFF
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -331,7 +331,7 @@ pub enum LexError {
     FloatUnderflow,
 
     #[error("{0}")]
-    InvalidHexFloat(hexponent::ParseError),
+    InvalidHexFloat(#[from] hexponent::ParseError),
 
     #[doc(hidden)]
     #[error("internal error: do not construct nonexhaustive variants")]
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -280,9 +280,6 @@ pub enum CppError {
 /// Lex errors are non-exhaustive and may have new variants added at any time
 #[derive(Clone, Debug, Error, PartialEq, Eq)]
 pub enum LexError {
-    #[error("{0}")]
-    Generic(String),
-
     #[error("unterminated /* comment")]
     UnterminatedComment,
 
@@ -332,6 +329,9 @@ pub enum LexError {
 
     #[error("underflow parsing floating literal")]
     FloatUnderflow,
+
+    #[error("{0}")]
+    InvalidHexFloat(hexponent::ParseError),
 
     #[doc(hidden)]
     #[error("internal error: do not construct nonexhaustive variants")]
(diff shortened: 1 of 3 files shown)


src/lex/mod.rs
```diff
@@ -366,8 +366,11 @@ impl Lexer {
             buf.push(c);
             self.next_char();
         }
-        let float = if hex {
-            hexf_parse::parse_hexf64(&buf, false).map_err(|err| err.to_string())
+        let float: f64 = if hex {
+            let float_literal: hexponent::FloatLiteral = buf
+                .parse()
+                .map_err(|err: hexponent::ParseError| err.to_string())?;
```
jyn514: Here's a good first improvement: add a specific error variant to `LexError` :)

src/lex/mod.rs
```diff
@@ -366,8 +366,11 @@ impl Lexer {
...
+            Ok(float_literal.into())
```
jyn514: It would be great to warn here if the conversion is imprecise.

src/data/error.rs
```diff
@@ -333,6 +330,9 @@ pub enum LexError {
     #[error("underflow parsing floating literal")]
     FloatUnderflow,
 
+    #[error("{0}")]
+    InvalidHexFloat(hexponent::ParseError),
```
jyn514: Can you use `#[from]` here? Or does ParseError not implement Error?

src/lex/mod.rs
```diff
@@ -376,10 +374,10 @@ impl Lexer {
...
-        let float = if hex {
-            // Leaving as generic because it will be replaced shortly.
-            hexf_parse::parse_hexf64(&buf, false)
-                .map_err(|err| LexError::Generic(err.to_string()))?
+        let float: f64 = if hex {
+            let float_literal: hexponent::FloatLiteral =
+                buf.parse().map_err(LexError::InvalidHexFloat)?;
+            float_literal.into()
         } else {
             buf.parse::<f64>()?
```
jyn514: I think you can remove this type annotation since you added one above.

src/data/error.rs
pythondude325: I can't use from because it requires an extra method not on `std::error::Error` that is implemented by the proc-macro.

jyn514: That doesn't match the documentation ...

> Any error type that implements std::error::Error or dereferences to dyn std::error::Error will work as a source.

https://docs.rs/crate/thiserror/1.0.9

src/lex/mod.rs
jyn514: As discussed on Discord, this needs a refactor so it can wait for a later PR.

src/lex/mod.rs
```diff
@@ -354,9 +354,7 @@ impl Lexer {
                 if !is_digit(self.peek()) {
                     // I'm leaving this as a generic error because I will get
                     // rid of it with hexponent.
```
jyn514: This comment should be removed.

PR #322, opened 2020-03-04
Add JIT

jyn514:
Same as #270 but with merge conflicts fixed.


src/lib.rs
```diff
@@ -232,6 +286,97 @@ pub fn link(obj_file: &Path, output: &Path) -> Result<(), io::Error> {
...
+        let (result, warnings) = compile::<SimpleJITBackend>(module, program, opt);
+        let result = result.map(JIT::from);
+        (result, warnings)
+    }
+
+    /// Invoke this function before trying to get access to "new" compiled functions.
+    pub fn finalize(&mut self) {
+        self.module.finalize_definitions();
+    }
+    /// Get compiled function, if this function doesn't exist then `None` is returned, otherwise it's address returned.
```
jyn514:
```suggestion
    /// Get compiled function, if this function doesn't exist then `None` is returned, otherwise its address returned.
```

src/lib.rs
```diff
@@ -232,6 +286,97 @@ pub fn link(obj_file: &Path, output: &Path) -> Result<(), io::Error> {
...
+        use cranelift_module::FuncOrDataId;
+
+        let name = self.module.get_name(name);
+        if let Some(FuncOrDataId::Func(id)) = name {
+            Some(self.module.get_finalized_function(id))
+        } else {
+            None
+        }
+    }
+    /// Get compiled static data, if this data doesn't exit then `None` is returned, otherwise it's andress and size returned.
```
jyn514:
```suggestion
    /// Get compiled static data. If this data doesn't exist then `None` is returned, otherwise its address and size are returned.
```

src/lib.rs
```diff
@@ -135,16 +139,67 @@ pub fn preprocess(
...
+        ObjectTrapCollection::Disabled,
+        cranelift_module::default_libcall_names(),
+    );
+    Module::new(builder)
+}
+
 /// Compile and return the declarations and warnings.
-pub fn compile(
+pub fn compile<B: Backend>(
+    module: Module<B>,
```
jyn514: Not sure I like this new API ... but it's fine for a first pass.

Issue #242, opened 2020-02-04
[ICE] crash on multiply in sizeof()


jyn514:
**Expected behavior**
RCC gives an error saying the array is too large.

**Actual Behavior**
RCC crashes with overflow.

**Code**

```c
char h[][3U-3827228678929574271][3] = {{4,5,6}};
The application panicked (crashed).
Message:  attempt to multiply with overflow
```

<details><summary>Backtrace</summary>

```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                          (6 post panic frames hidden)                          
 6: core::panicking::panic::ha556f2c3753f3759
    at src/libcore/panicking.rs:70
 7: <u64 as core::ops::arith::Mul>::mul::h5181c21b5c178bc2
    at /rustc/73528e339aae0f17a15ffa49a8ac608f50c6cf14/src/libcore/ops/arith.rs:318
 8: <u64 as core::ops::arith::Mul<&u64>>::mul::h1d1276c4c74de715
    at src/lib.rs:1
 9: rcc::arch::<impl rcc::data::types::Type>::sizeof::{{closure}}::hed6dc419ce36a058
    at src/arch/mod.rs:74
10: core::result::Result<T,E>::and_then::haa7997835743a121
    at /rustc/73528e339aae0f17a15ffa49a8ac608f50c6cf14/src/libcore/result.rs:691
11: rcc::arch::<impl rcc::data::types::Type>::sizeof::h82206b81eeb141c1
    at src/arch/mod.rs:74
12: rcc::arch::<impl rcc::data::types::Type>::sizeof::h82206b81eeb141c1
    at src/arch/mod.rs:74
13: rcc::ir::static_init::<impl rcc::ir::Compiler>::store_static::had90360549edd56e
    at src/ir/static_init.rs:77
14: rcc::ir::compile::haeb8251a574ff20c
    at src/ir/mod.rs:89
15: rcc::compile::h87633814d689a39f
    at src/lib.rs:98
16: rcc::real_main::h1dd2cb989e6fa0d8
    at src/main.rs:101
17: rcc::main::h396b827397690aa0
    at src/main.rs:169
                        (10 runtime init frames hidden)
```

</details>

jyn514:
Instructions:

[...]

Issue #156, opened 2019-12-15
Error system mega-issue


pythondude325:
I have decided to write an issue to track all the changes I am making to the error system and tasks required for these changes.

Tasks:
- [x] New error trait (In progress #154)
- [ ] Change warnings to use the new error system (#114)
- [ ] Replace old error printing with new system
- [ ] Better `define_error!` ergonomics
- [ ] Replace old error type usage
- [ ] Create more specific error types
- [ ] Get rid of all uses of the "Generic" errors
- [ ] Change tests to check that the right error was returned
- [ ] Create a new error handling system. (Error handler queue instead of errors in pending)

I plan on working on all of these but I am willing to have other people do them too.

jyn514:
Idea for warnings: return a newtype in Iterator that looks like this:

```rust
pub enum CompileItem {
    Ok(Declaration),
    Warn(CompileWarning),
    Err(CompileErr),
}
```

but still keep `CompileResult` all through the rest of the codebase. Once that's done, we can rename `CompileItem` -> `CompileResult` and `CompileResult` -> `CompileResultInternal` or something

jyn514:
Mentoring instructions:

If you want to switch from a `Generic` error to a new specific error, write a patch that looks like this: https://github.com/jyn514/rcc/pull/192

Issue #63, opened 2019-09-29
Variable Length Arrays


jyn514:
Thanks to @Tachyon

```c
int main() {
        int n = 1;
        int a[n];
        return *a;
}
<stdin>:3:9: error: not a constant expression
```

jyn514:
It may be more elegant to first implement `alloca` as a builtin and then implement this in terms of `alloca`.

xTachyon:
Idea - always allocate the same space for a vla and at runtime check if you have enough static allocated space, yay, otherwise crash or something.

It's not ideal, but it's better than nothing.

Remind @jyn514 in 2 months to check for cranelift support.

PR #325, opened 2020-03-05
proptest for alignof() and sizeof() of Type

Byter09:
For now its only minimal recursion as it is pretty tricky to get this right, but maybe someone wants to help out :)


src/arch/mod.rs
```diff
@@ -10,14 +10,16 @@ use cranelift::codegen::{
     },
     isa::{CallConv, TargetIsa},
 };
-use lazy_static::lazy_static;
 use target_lexicon::Triple;
 
+use lazy_static::lazy_static;
+pub(crate) use x64::*;
+use Type::*;
+
```
jyn514: Please don't reorder these imports.

src/arch/mod.rs
```diff
@@ -344,4 +350,18 @@ mod tests {
         let ty = struct_for_types(vec![Pointer(Box::new(Int(true))), Int(true)]);
         assert_eq!(ty.alignof(), Ok(8));
     }
+
+    proptest! {
+        #[test]
+        fn proptest_sizeof_alignof_same_variant(t in arb_type()) {
```
jyn514: This can just be `sizeof_alignof` since we're testing things besides the discriminant

src/data/types.rs
```diff
@@ -386,3 +392,37 @@ impl FunctionType {
...
+        let leaf = prop_oneof![
+            Just(Type::Void),
+            Just(Type::Bool),
+            any::<bool>().prop_map(Type::Char),
+            any::<bool>().prop_map(Type::Short),
+            any::<bool>().prop_map(Type::Int),
+            any::<bool>().prop_map(Type::Long),
+            Just(Type::Float),
+            Just(Type::Double),
+            //Type::Enum(Option<InternedStr>, Vec<(InternedStr, i64)>),
```
jyn514: I thought you implemented arbitrary for InternedStr? Is there something else making this tricky?

src/data/types.rs
```diff
@@ -386,3 +392,37 @@ impl FunctionType {
...
+            Just(Type::Error),
+        ];
+
+        leaf.prop_recursive(8, 256, 10, |inner| {
+            prop_oneof![
+                inner.prop_map(|t| Type::Pointer(Box::new(t))),
+                //Type::Function(FunctionType),
+                //Type::Union(StructType),
+                //Type::Struct(StructType),
+                //Type::Array(Box<Type>, ArrayType),
```
jyn514: This should be basically the same as Pointer since ArrayType isn't recursive.

src/data/types.rs
```diff
@@ -386,3 +392,37 @@ impl FunctionType {
...
+                //Type::Bitfield(Vec<BitfieldType>),
```
jyn514: This can be removed, if/when I implement bitfields they will be annotations on StructType, not a separate variant.

src/data/types.rs
```diff
@@ -386,3 +392,37 @@ impl FunctionType {
...
+            Just(Type::VaList),
+            Just(Type::Error),
+        ];
+
+        leaf.prop_recursive(8, 256, 10, |inner| {
+            prop_oneof![
+                inner.prop_map(|t| Type::Pointer(Box::new(t))),
+                //Type::Function(FunctionType),
+                //Type::Union(StructType),
+                //Type::Struct(StructType),
```
jyn514: I can help with StructType this afternoon.

src/data/types.rs
Byter09: No, there is a function providing an arbitrary one, but not a direct trait implementation, as that is kind of tricky. If you want to give it a go though, have at it! :D

src/arch/mod.rs
Byter09: Forgot to remove "Reorganize Imports" in CLion. It sorts them by scope though, don't know what's bad about it?

jyn514: I guess it's not a big deal but it distracts from the actual changes. We can keep these ones.

Byter09:
Tests will no fail because it instantly detects this: https://github.com/jyn514/rcc/issues/242

Fixing that using https://github.com/jyn514/rcc/issues/156 requires changes to function return types, which causes a ripple effect, ending in functions I have no idea how to convert the errors to. So... kinda blocked by that. Easy fix would be a simple `Err("sizeof overflow")`

jyn514:
Let's do the simple fix for now and we can come back to refactoring `sizeof()`.

Byter09:
Can I somehow mark this PR as blocked by https://github.com/jyn514/rcc/pull/327?

jyn514:
The CI test is failing because the `sizeof() == alignof()` test is incorrect - if you have an unbounded array, you can take the alignment even without knowing the size. This won't change until VLAs are implemented (#63), which will require a significant refactor of `sizeof()` anyway. In the meantime, I think the test should be removed.

PR #333, opened 2020-03-07
Don't error on extern redeclarations of static functions

### LINKED ISSUE #311: Static functions redeclared as extern are static
**Expected behavior**
The second `extern int f();` should have no effect.

**Actual Behavior**
rcc gives a type error.

**Code**

```c
static int f();
extern int f();
<stdin>:2:13 error: invalid program: redeclaration of 'f' with different type or qualifiers (originally static int f(), now extern int f())
```

http://port70.net/~nsz/c/c11/n1570.html#6.2.2p4:

> 4 For an identifier declared with the storage-class specifier extern in a scope in which a prior declaration of that identifier is visible,31) if the prior declaration specifies internal or external linkage, the linkage of the identifier at the later declaration is the same as the linkage specified at the prior declaration. If no prior declaration is visible, or if the prior declaration specifies no linkage, then the identifier has external linkage.

repnop:
Fixes #311

### DIFF
--- a/src/parse/decl.rs
+++ b/src/parse/decl.rs
@@ -271,11 +271,16 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
             decl.storage_class = StorageClass::Auto;
         }
         if let Some(existing) = self.scope.get_immediate(&decl.id) {
+            let extern_redecl_of_static = match (existing.storage_class, decl.storage_class) {
+                (StorageClass::Static, StorageClass::Extern) => true,
+                _ => false,
+            };
+
             if existing == decl {
                 if decl.init && existing.init {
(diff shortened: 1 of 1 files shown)


src/parse/decl.rs
```diff
@@ -271,11 +271,16 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
             decl.storage_class = StorageClass::Auto;
         }
         if let Some(existing) = self.scope.get_immediate(&decl.id) {
+            let extern_redecl_of_static = match (existing.storage_class, decl.storage_class) {
```
jyn514: Do you need a match here? I think it would be simpler to use `==`.

repnop: it's a pretty long comparison that rustfmt broke up onto a new line, match keeps it more readable imo

Issue #265, opened 2020-02-09
[ICE] Panic on any two-byte utf-8 character


pythondude325:
**Expected behavior**
An error saying that the character is not allowed

**Actual Behavior**
RCC panics when it finds the character

**Code**
```c
χ
```

<details><summary>Backtrace</summary>

```
The application panicked (crashed).
Message:  end location should be in bounds: InvalidCharBoundary { given: ByteIndex(1) }
Location: src/libcore/result.rs:1165

Run with RUST_BACKTRACE=full to include source snippets.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                          (7 post panic frames hidden)
 7: core::result::Result<T,E>::expect::h1d1a93e7528a9764
    at /rustc/73528e339aae0f17a15ffa49a8ac608f50c6cf14/src/libcore/result.rs:960
 8: rcc::pretty_print::h6c7743c73b92b0aa
    at src/main.rs:278
 9: rcc::error::h37a51aebaa39649c
    at src/main.rs:245
10: rcc::err_exit::h88d58006c51ac450
    at src/main.rs:208
11: rcc::main::{{closure}}::hd6391ab094d459b3
    at src/main.rs:136
12: core::result::Result<T,E>::unwrap_or_else::h14bef5d2a34cff97
    at /rustc/73528e339aae0f17a15ffa49a8ac608f50c6cf14/src/libcore/result.rs:818
13: rcc::main::h93a3ad722c94810d
    at src/main.rs:135
                        (12 runtime init frames hidden)
```

</details>

jyn514:
Regression from #260.

jyn514:
Might be easier to just switch to codespan-reporting?

jyn514:
I think this is unrelated to codespan-reporting/the pretty printer, it's because the lexer is working over `u8` and immediately gives an error instead of consuming the whole unicode character.

jyn514:
I think I should just revert #260, even rustc uses `.chars()`: https://github.com/rust-lang/rust/blob/master/src/librustc_lexer/src/lib.rs#L280

PR #335, opened 2020-03-08
Removed (almost) all generic lex errors

pythondude325:
I left a few generic errors in because I will get rid of them in the hexponent PR.

### DIFF
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -275,6 +275,43 @@ pub enum CppError {
     __Nonexhaustive,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum Radix {
+    Binary,
+    Octal,
+    Decimal,
+    Hexadecimal,
+}
+
+use std::fmt;
+impl fmt::Display for Radix {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                Radix::Binary => "binary",
+                Radix::Octal => "octal",
+                Radix::Decimal => "decimal",
+                Radix::Hexadecimal => "hexadecimal",
+            }
+        )
+    }
+}
+
+impl std::convert::TryFrom<u32> for Radix {
+    type Error = ();
+    fn try_from(int: u32) -> Result<Radix, ()> {
+        match int {
+            2 => Ok(Radix::Binary),
+            8 => Ok(Radix::Octal),
+            10 => Ok(Radix::Decimal),
+            16 => Ok(Radix::Hexadecimal),
+            _ => Err(()),
+        }
+    }
+}
+
 /// Lex errors are non-exhaustive and may have new variants added at any time
 #[derive(Clone, Debug, Error, PartialEq, Eq)]
 pub enum LexError {
@@ -286,6 +323,45 @@ pub enum LexError {
 
     #[error("no newline at end of file")]
     NoNewlineAtEOF,
+
+    #[error("unknown token: {0:?}")]
+    UnknownToken(u8),
+
+    #[error("missing terminating \" character in string literal")]
+    StringMissingEndQuote,
+
+    #[error("illegal newline while parsing string literal")]
+    NewlineInString,
+
+    #[error("{0} character escape out of range")]
+    InvalidNumericCharEscape(Radix),
+
+    #[error("overflow while parsing integer literal")]
+    IntegerOverflow,
+
+    #[error("exponent for floating literal has no digits")]
+    FloatExponentMissingDigits,
+
+    #[error("missing digits to {0} integer constant")]
+    MissingDigits(Radix),
+
+    #[error("{0}")]
+    ParseFloat(#[from] std::num::ParseFloatError),
+
+    #[error("invalid digit {digit} in {radix} constant")]
+    InvalidDigit { digit: u32, radix: Radix },
+
+    #[error("multi-character character literal")]
+    MultiCharCharLiteral,
+
+    #[error("missing terminating ' character in char literal")]
+    CharMissingEndQuote,
+
+    #[error("illegal newline while parsing char literal")]
+    NewlineInChar,
+
+    #[error("empty character constant")]
+    EmptyChar,
 
     #[doc(hidden)]
     #[error("internal error: do not construct nonexhaustive variants")]
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -293,7 +293,7 @@ pub enum LexError {
     UnknownToken(char),
 
     #[error("missing terminating {} character in {} literal",
-        if *(.string) { "\"" } else { "\'" },
+        if *(.string) { "\"" } else { "'" },
         if *(.string) { "string" } else { "character" })]
     MissingEndQuote { string: bool },
 
@@ -303,8 +303,11 @@ pub enum LexError {
     #[error("{0} character escape out of range")]
     CharEscapeOutOfRange(Radix),
 
-    #[error("overflow while parsing integer literal")]
-    IntegerOverflow,
+    #[error("overflow while parsing {}integer literal",
+        if let &Some(signed) = .is_signed {
+            if signed { "signed "} else { "unsigned "}
+        } else { "" })]
+    IntegerOverflow { is_signed: Option<bool> },
 
     #[error("exponent for floating literal has no digits")]
     ExponentMissingDigits,
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -304,8 +304,8 @@ pub enum LexError {
     CharEscapeOutOfRange(Radix),
 
     #[error("overflow while parsing {}integer literal",
-        if let &Some(signed) = .is_signed {
-            if signed { "signed "} else { "unsigned "}
+        if let Some(signed) = .is_signed {
+            if *signed { "signed "} else { "unsigned "}
         } else { "" })]
     IntegerOverflow { is_signed: Option<bool> },
 
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -286,16 +286,13 @@ pub enum Radix {
 use std::fmt;
 impl fmt::Display for Radix {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(
-            f,
-            "{}",
-            match self {
-                Radix::Binary => "binary",
-                Radix::Octal => "octal",
-                Radix::Decimal => "decimal",
-                Radix::Hexadecimal => "hexadecimal",
-            }
-        )
+        let word = match self {
+            Radix::Binary => "binary",
+            Radix::Octal => "octal",
+            Radix::Decimal => "decimal",
+            Radix::Hexadecimal => "hexadecimal",
+        };
+        write!(f, "{}", word)
     }
 }
 
@@ -327,8 +324,10 @@ pub enum LexError {
     #[error("unknown token: {0:?}")]
     UnknownToken(u8),
 
-    #[error("missing terminating \" character in string literal")]
-    StringMissingEndQuote,
+    #[error("missing terminating {} character in {} literal",
+        if *(.string) { "\"" } else { "\'" },
+        if *(.string) { "string" } else { "character" })]
+    MissingEndQuote { string: bool },
 
     #[error("illegal newline while parsing string literal")]
     NewlineInString,
@@ -340,7 +339,7 @@ pub enum LexError {
     IntegerOverflow,
 
     #[error("exponent for floating literal has no digits")]
-    FloatExponentMissingDigits,
+    ExponentMissingDigits,
 
     #[error("missing digits to {0} integer constant")]
     MissingDigits(Radix),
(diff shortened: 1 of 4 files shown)


src/data/error.rs
```diff
@@ -275,6 +275,43 @@ pub enum CppError {
...
+    Binary,
+    Octal,
+    Decimal,
+    Hexadecimal,
+}
+
+use std::fmt;
+impl fmt::Display for Radix {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
```
jyn514: I don't like how rustfmt formatted this, maybe assign to a temporary variable first?

src/data/error.rs
```diff
@@ -287,6 +324,45 @@ pub enum LexError {
     #[error("no newline at end of file")]
     NoNewlineAtEOF,
 
+    #[error("unknown token: {0:?}")]
+    UnknownToken(u8),
```
jyn514: Can you print this as a `char` instead? I know this isn't unicode safe but the current output isn't very helpful:

```
`
<stdin>:1:1 error: invalid token: unknown token 96
```

src/lex/mod.rs
```diff
@@ -893,7 +868,6 @@ impl Iterator for Lexer {
             }
         }
         // oof
```
jyn514: I think this comment can be deleted now :)

src/lex/tests.rs
```diff
@@ -171,8 +171,6 @@ fn test_float_literals() {
 fn test_num_errors() {
     assert_err("1e");
     assert_err("1e.");
-    assert_err("1e100000");
-    assert_err("1e-100000");
```
jyn514: Why were these tests removed?

src/data/error.rs
```diff
@@ -287,6 +324,45 @@ pub enum LexError {
...
+
+    #[error("missing terminating \" character in string literal")]
+    StringMissingEndQuote,
+
+    #[error("illegal newline while parsing string literal")]
+    NewlineInString,
+
+    #[error("{0} character escape out of range")]
+    InvalidNumericCharEscape(Radix),
```
jyn514: Maybe `CharEscapeOutOfRange` would be better? `InvalidNumeric` doesn't tell you why it was better. `OutOfRange` seems a little verbose though ... I'm open to suggestions.

src/data/error.rs
```diff
@@ -287,6 +324,45 @@ pub enum LexError {
     #[error("no newline at end of file")]
     NoNewlineAtEOF,
 
+    #[error("unknown token: {0:?}")]
+    UnknownToken(u8),
+
+    #[error("missing terminating \" character in string literal")]
+    StringMissingEndQuote,
```
jyn514: Do you want to combine this into `MissingEndQuote { string: bool }`? I think that would make it easier to create at the call site too.

src/lex/mod.rs
```diff
@@ -279,10 +279,7 @@ impl Lexer {
                 if radix == 8 || radix == 10 || self.peek() == Some(b'.') {
                     start
                 } else {
-                    return Err(format!(
-                        "missing digits to {} integer constant",
-                        if radix == 2 { "binary" } else { "hexadecimal" }
-                    ));
+                    return Err(LexError::MissingDigits(radix.try_into().unwrap()));
```
jyn514: Could you make `radix` a `Radix` instead of storing it as a u8? Then we wouldn't have this `unwrap()`. You might have to make a `.as_u8()` function for when the lexer multiplies by the radix.

src/lex/mod.rs
```diff
@@ -296,12 +293,10 @@ impl Lexer {
             return float.map(float_literal);
         }
         let literal = if self.match_next(b'u') || self.match_next(b'U') {
-            let unsigned = u64::try_from(digits)
-                .map_err(|_| "overflow while parsing unsigned integer literal")?;
+            let unsigned = u64::try_from(digits).map_err(|_| LexError::IntegerOverflow)?;
```
jyn514: Can you keep the information about whether it was signed or unsigned?

src/lex/mod.rs
```diff
@@ -353,14 +348,18 @@ impl Lexer {
...
+                        "exponent for floating literal has no digits".to_string(),
+                    ));
                 }
                 buf.push('p');
                 buf.push(self.next_char().unwrap() as char);
             }
         } else if self.match_next(b'e') || self.match_next(b'E') {
             if !is_digit(self.peek()) {
-                return Err(String::from("exponent for floating literal has no digits"));
+                return Err(LexError::FloatExponentMissingDigits);
```
jyn514: I think this can just be `ExponentMissingDigits`, having an exponent automatically makes it a float.

src/data/error.rs
```diff
@@ -287,6 +324,45 @@ pub enum LexError {
...
+    IntegerOverflow,
+
+    #[error("exponent for floating literal has no digits")]
+    FloatExponentMissingDigits,
+
+    #[error("missing digits to {0} integer constant")]
+    MissingDigits(Radix),
+
+    #[error("{0}")]
+    ParseFloat(#[from] std::num::ParseFloatError),
```
jyn514: Oh this is clever! I wouldn't have thought of this :)

src/lex/mod.rs
```diff
@@ -374,31 +373,21 @@ impl Lexer {
...
-        }?;
-        if float.is_infinite() {
-            return Err("overflow parsing floating literal".into());
-        }
-        let should_be_zero = buf.bytes().all(|c| match c {
-            b'.' | b'+' | b'-' | b'e' | b'p' | b'0' => true,
-            _ => false,
-        });
-        if float == 0.0 && !should_be_zero {
-            Err("underflow parsing floating literal".into())
```
jyn514: I know we decided to get rid of the check for `INF` - is there a reason you also got rid of the check for underflow?

src/lex/mod.rs
```diff
@@ -374,31 +373,21 @@ impl Lexer {
             self.next_char();
         }
         let float = if hex {
```
jyn514: You can avoid this assignment + `?` conversion by having the `if` be the last expression in the block:

```rust
if hex {
    parse_hexf64().map_err()
} else {
    buf.parse()
}

File: src/lex/mod.rs [hunk 13]
```diff
@@ -374,31 +373,21 @@ impl Lexer {
...
-        }
+            buf.parse::<f64>()?
+        };
+        Ok(float)
     }
     // returns None if there are no digits at the current position
     fn parse_int(
         &mut self,
         mut acc: u64,
         radix: u32,
```
jyn514: Same as above, it would be nice to take a `Radix` instead of a u32 to avoid the unwrap.

src/lex/mod.rs
```diff
@@ -568,10 +550,8 @@ impl Lexer {
             }
         }
         let (term_err, newline_err) = (
-            Err(String::from(
-                "Missing terminating ' character in char literal",
-            )),
-            Err(String::from("Illegal newline while parsing char literal")),
+            Err(LexError::CharMissingEndQuote),
+            Err(LexError::NewlineInChar),
```
jyn514: I think you can get rid of this assignment in favor of just using the variants directly. I only had this here because the error message was long and I didn't want to type it more than once lol

src/data/error.rs
pythondude325: How about a write macro in each branch of the match?

src/data/error.rs
pythondude325: What should I do when it's not unicode valid? Panicing might come back to bite us.

src/data/error.rs
jyn514: Sure, that works too, it would just be a little more code.

src/lex/tests.rs
pythondude325: These tests made sure that errors were raised when floats overflowed/underflowed.

src/data/error.rs
pythondude325: I would have chosen `InvalidCharEscape`, but that makes it seem like it might be an incorrect letter after a slash in a char literal (ex: `\j`)

src/data/error.rs
pythondude325: I chose to do separate errors because the error directive would have been large. I can still do it though.

src/lex/mod.rs
pythondude325: Should I move it out of the `error` module then?

src/lex/mod.rs
pythondude325: The problem is that this same variant is also used in a place where the sign is not known. I could make it an `Option<bool>` if you like.

src/data/error.rs
pythondude325: I plan to do this with hexponent's error type too.

src/lex/mod.rs
pythondude325: Can't because of the underflow check.

src/data/error.rs
pythondude325: Maybe try converting to Unicode and print as hex char literal if it can't.

jyn514: This already panics: https://github.com/jyn514/rcc/issues/265. Casting to `char` can't panic anyway since it will turn it into the appropriate unicode scalar value.

src/lex/mod.rs
jyn514: Yes, please do.

src/data/error.rs
jyn514: Right, my point is that for `InvalidNumericCharEscape` to make sense, you have to already know the only thing that can make it invalid is an escape that's out of range.

src/lex/mod.rs
jyn514: I think the `Option<bool>` would be a good idea, yeah.

src/data/error.rs
```diff
@@ -327,8 +324,10 @@ pub enum LexError {
     #[error("unknown token: {0:?}")]
     UnknownToken(u8),
 
-    #[error("missing terminating \" character in string literal")]
-    StringMissingEndQuote,
+    #[error("missing terminating {} character in {} literal",
+        if *(.string) { "\"" } else { "\'" },
```
jyn514:
```suggestion
        if *(.string) { "\"" } else { "'" },
```

src/lex/mod.rs
jyn514: Ah right, mutually exclusive review comments :laughing: 

offtopic: `rustfix` actually had this exact same issue: https://discordapp.com/channels/442252698964721669/446773253953028107/651615006898782233

src/data/error.rs
pythondude325: Okay, I will just rename it to `CharEscapeOutOfRange` then.

src/lex/mod.rs
```diff
@@ -305,20 +305,20 @@ impl Lexer {
...
     }
     // at this point we've already seen a '.', if we see one again it's an error
-    fn parse_float(&mut self, radix: u32, mut buf: String) -> Result<f64, LexError> {
+    fn parse_float(&mut self, radix: Radix, mut buf: String) -> Result<f64, LexError> {
         buf.push('.');
         // parse fraction: second {digits} in regex
         while let Some(c) = self.peek() {
             let c = c as char;
-            if c.is_digit(radix) {
+            if c.is_digit(radix.as_u8() as u32) {
```
jyn514: Use `into()` instead of `as` for non-lossy conversions.
```suggestion
            if c.is_digit(radix.as_u8().into()) {
```

src/lex/mod.rs
```diff
@@ -393,21 +397,18 @@ impl Lexer {
...
         &mut self,
         mut acc: u64,
-        radix: u32,
+        radix: Radix,
         buf: &mut String,
     ) -> Result<Option<u64>, LexError> {
         let parse_digit = |c: char| match c.to_digit(16) {
             None => Ok(None),
-            Some(digit) if digit < radix => Ok(Some(digit)),
+            Some(digit) if digit < (radix.as_u8() as u32) => Ok(Some(digit)),
```
jyn514:
```suggestion
            Some(digit) if digit < (radix.as_u8().into()) => Ok(Some(digit)),
```

src/lex/mod.rs
```diff
@@ -431,15 +432,15 @@ impl Lexer {
             };
             buf.push(c as char);
             let maybe_digits = acc
-                .checked_mul(radix.into())
+                .checked_mul(radix.as_u8() as u64)
```
jyn514:
```suggestion
                .checked_mul(radix.as_u8().into())
```

src/lex/mod.rs
```diff
@@ -848,7 +849,9 @@ impl Iterator for Lexer {
                     }
                 }
                 x => {
-                    return Some(Err(self.span(span_start).with(LexError::UnknownToken(x))));
+                    return Some(Err(self
+                        .span(span_start)
+                        .with(LexError::UnknownToken(x as char))));
```
jyn514: I was imagining the `as char` cast in the Display impl ... but I guess it doesn't really matter.

pythondude325: I think the errors should be as transparent as possible

PR #365, opened 2020-04-07
Fix #142

### LINKED ISSUE #142: int() should decay to int(*)() in declarations

hdamron17:
Decay parameter of 'function returning type' to 'pointer to function returning type'

### DIFF
--- a/src/parse/decl.rs
+++ b/src/parse/decl.rs
@@ -948,10 +948,10 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
                     ctype = Type::Pointer(to);
                 }
 
-                // Fix #142 (C11 Standard 6.7.6.3 paragraph 8)
+                // C11 Standard 6.7.6.3 paragraph 8
                 // "A declaration of a parameter as 'function returning type' shall be
                 //  adjusted to 'pointer to function returning type', as in 6.3.2.1."
-                if let Type::Function(_) = ctype {
+                if ctype.is_function() {
                     ctype = Type::Pointer(Box::new(ctype));
                 }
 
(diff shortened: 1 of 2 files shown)


hdamron17:
* Let me know if there is a better way to pattern match `ctype` as a `Type::Function`.
* I don't know the implications of creating a new Box (if any) since I am not a true Rustacean.
* I added a test C code, but should I add any unit test for something like this? The C code referenced in #142 segfaults because it calls `f()` with the wrong number of parameters so I used a more sensible test case.

jyn514:
> Let me know if there is a better way to pattern match ctype as a Type::Function.

You can also use `is_function()` which I implemented on `Type`, but the way you have it now is fine.

> I don't know the implications of creating a new Box (if any) since I am not a true Rustacean.

This is fine (and actually required by the `Type` enum because it's a recursive type). All it does is move the struct to the heap, sort of like `new` in C++.

> I added a test C code, but should I add any unit test for something like this? The C code referenced in #142 segfaults because it calls f() with the wrong number of parameters so I used a more sensible test case.

The unit tests are currently [pretty ugly](https://github.com/jyn514/rcc/blob/master/src/parse/decl.rs#L2135) so I think a runner test is fine. The segfault in 142 is because it calls `f()` infinitely many times and runs out of stack space.

src/parse/decl.rs
```diff
@@ -947,6 +947,14 @@ impl<I: Iterator<Item = Lexeme>> Parser<I> {
                 if let Type::Array(to, _) = ctype {
                     ctype = Type::Pointer(to);
                 }
+
+                // Fix #142 (C11 Standard 6.7.6.3 paragraph 8)
```
jyn514, line 951: Normally I put a link to the issue in the test case, not in the code, unless it's something really hacky. The link to the standard is good, though.

hdamron17: Sounds good. I'll change to use `is_function()` while I'm making this change anyway.

PR #373, opened 2020-04-21
Tell rustfmt to make diffs smaller

jyn514:


jyn514:
A little birdy has informed me this is not a popular change.

PR #377, opened 2020-04-21
Fix travis builds

jyn514:
No longer require compiles to be done in a git repo

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -7,7 +7,7 @@ fn write_git_version() {
         let commit = git(&["rev-parse", "--short=7", "HEAD"]).unwrap_or_else(|| "????".into());
         let most_recent_tag = match git(&["describe", "--tags", "--abbrev=0", "HEAD"]) {
             Some(c) => c,
-            None => return "????".into(),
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -4,9 +4,11 @@ fn main() {
 fn write_git_version() {
     let tag = git(&["describe", "--exact-match", "--tags", "HEAD"]);
     let version = tag.unwrap_or_else(|| {
-        let commit = git(&["rev-parse", "--short=7", "HEAD"])
-            .expect("`git` was not successful, are you in the right directory?");
-        let most_recent_tag = git(&["describe", "--tags", "--abbrev=0", "HEAD"]).unwrap();
+        let commit = git(&["rev-parse", "--short=7", "HEAD"]).unwrap_or_else(|| "????".into());
+        let most_recent_tag = match git(&["describe", "--tags", "--abbrev=0", "HEAD"]) {
+            Some(c) => c,
+            None => return "????".into(),
```
jyn514, line 10: This should probably return the `CARGO_PACKAGE_VERSION` instead

PR #381, opened 2020-04-25
Update cranelift to 0.63

jyn514:
~~I'm not super happy with this, I had to add a new dependency on `object` that's not linked to `cranelift_object`, so it will break when cranelift_object updates.~~ Resolved in 0.63

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ jobs:
               bash <(curl -s https://codecov.io/bash) &&
               echo "Uploaded code coverage"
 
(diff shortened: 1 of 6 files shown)


jyn514:
Waiting on https://github.com/bytecodealliance/wasmtime/pull/1599

jyn514:
Note that since object uses `ptr.cast()` this breaks compatibility with 1.37. The MSRV is now 1.38.

jyn514:
The failing check is because travis has an old version of `beta`, stable passes without a problem.

PR #403, opened 2020-05-03
Color flag

pythondude325:
Add the standard `--color={always,auto,never}` CLI option. I might have missed some places where color is used.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,6 +14,9 @@ extern crate color_backtrace;
 extern crate pico_args;
 extern crate rcc;
 
+#[cfg(debug_assertions)]
+use color_backtrace::termcolor;
+
 use ansi_term::{ANSIString, Colour};
 use codespan::FileId;
 use pico_args::Arguments;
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -29,12 +32,20 @@ static ERRORS: AtomicUsize = AtomicUsize::new(0);
 static WARNINGS: AtomicUsize = AtomicUsize::new(0);
 
 const HELP: &str = concat!(
-    env!("CARGO_PKG_NAME"), " ", env!("RCC_GIT_REV"), "\n",
+    env!("CARGO_PKG_NAME"),
+    " ",
+    env!("RCC_GIT_REV"),
+    "\n",
```
jyn514, line 38: This makes it look less like the output. Before one line of source corresponded to one line of help output.

src/main.rs
```diff
@@ -62,7 +73,8 @@ OPTIONS:
 
 ARGS:
     <file>    The file to read C source from. \"-\" means stdin (use ./- to read a file called '-').
```
jyn514, line 75: Because you added `r#"`, this now shows a literal backslash.

src/main.rs
```diff
@@ -79,6 +91,47 @@ struct BinOpt {
...
+        match s {
+            "always" => Ok(ColorChoice::Always),
+            "auto" => Ok(ColorChoice::Auto),
+            "never" => Ok(ColorChoice::Never),
+            _ => Err("Invalid color choice"),
+        }
+    }
+}
+
+impl Into<color_backtrace::termcolor::ColorChoice> for ColorChoice {
```
jyn514, line 127:
```suggestion
impl Into<termcolor::ColorChoice> for ColorChoice {
```

src/main.rs
```diff
@@ -79,6 +91,47 @@ struct BinOpt {
...
+    fn into(self) -> termcolor::ColorChoice {
+        match self {
+            ColorChoice::Always => termcolor::ColorChoice::Always,
+            ColorChoice::Auto => termcolor::ColorChoice::Auto,
+            ColorChoice::Never => termcolor::ColorChoice::Never,
```
jyn514, line 132: Personally I would `use termcolor::ColorChoice::*;` here, but it's up to you.

src/main.rs
```diff
@@ -177,6 +232,16 @@ fn main() {
...
     };
+
+    #[cfg(debug_assertions)]
+    {
+        color_backtrace::install_with_settings(color_backtrace::Settings::new().output_stream(
+            Box::new(color_backtrace::termcolor::StandardStream::stderr(
+                opt.color.into(),
+            )),
+        ));
+    }
```
jyn514, line 243: Is the outer block still necessary? It looks like you removed the `if`.

src/main.rs
pythondude325: I think rustfmt just did that. A little odd

src/main.rs
pythondude325: I usually prefer having paths

PR #406, opened 2020-05-05
add dustmite based input minimizer

Byter09:
Before approving this MR, please play around a bit with the script. I want to make this as easy as possible to use.

### DIFF
--- a/minimizer/conditions/output_contains.sh
+++ b/minimizer/conditions/output_contains.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+cargo run -- --color never input.c 2>&1 | grep "$1"
+exit $?;
+ 
--- a/minimizer/conditions/output_contains.sh
+++ b/minimizer/conditions/output_contains.sh
@@ -1,5 +1,2 @@
-#!/usr/bin/env bash
-
+#!/bin/sh
 cargo run -- --color never input.c 2>&1 | grep "$1"
-exit $?;
- 
--- a/minimizer/conditions/output_contains.sh
+++ b/minimizer/conditions/output_contains.sh
@@ -1,2 +1,2 @@
 #!/bin/sh
-cargo run -- --color never input.c 2>&1 | grep "$1"
+source ./../run_cargo.sh | grep "$1"
--- a/minimizer/conditions/output_contains.sh
+++ b/minimizer/conditions/output_contains.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 # This is being executed inside of `rcc/minimizer/workspace/input.test` and thus requires `..` twice.
-source ./../../run_cargo.sh | grep "$1"
+. ./../../run_cargo.sh | grep "$1"
--- a/minimizer/conditions/output_contains.sh
+++ b/minimizer/conditions/output_contains.sh
@@ -1,2 +1,3 @@
 #!/bin/sh
-source ./../run_cargo.sh | grep "$1"
+# This is being executed inside of `rcc/minimizer/workspace/input.test` and thus requires `..` twice.
+source ./../../run_cargo.sh | grep "$1"
--- a/minimizer/conditions/return_code_equals.sh
+++ b/minimizer/conditions/return_code_equals.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+cargo run -- --color never input.c 2>&1;
+if [[ $? == $1 ]]; then exit 0; fi
+exit 1;
--- a/minimizer/conditions/return_code_equals.sh
+++ b/minimizer/conditions/return_code_equals.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
-cargo run -- --color never input.c 2>&1;
+source ./../run_cargo.sh
 [ $? = "$1" ]
--- a/minimizer/conditions/return_code_equals.sh
+++ b/minimizer/conditions/return_code_equals.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 # This is being executed inside of `rcc/minimizer/workspace/input.test` and thus requires `..` twice.
-source ./../../run_cargo.sh
+. ./../../run_cargo.sh
 [ $? = "$1" ]
--- a/minimizer/conditions/return_code_equals.sh
+++ b/minimizer/conditions/return_code_equals.sh
@@ -1,5 +1,3 @@
-#!/usr/bin/env bash
-
+#!/bin/sh
 cargo run -- --color never input.c 2>&1;
-if [[ $? == $1 ]]; then exit 0; fi
-exit 1;
+[ $? = "$1" ]
--- a/minimizer/conditions/return_code_equals.sh
+++ b/minimizer/conditions/return_code_equals.sh
@@ -1,3 +1,4 @@
 #!/bin/sh
-source ./../run_cargo.sh
+# This is being executed inside of `rcc/minimizer/workspace/input.test` and thus requires `..` twice.
+source ./../../run_cargo.sh
 [ $? = "$1" ]
--- a/minimizer/conditions/return_code_is_not.sh
+++ b/minimizer/conditions/return_code_is_not.sh
@@ -1,3 +1,4 @@
 #!/bin/sh
-source ./../run_cargo.sh
+# This is being executed inside of `rcc/minimizer/workspace/input.test` and thus requires `..` twice.
+source ./../../run_cargo.sh
 ! [ $? = "$1" ]
--- a/minimizer/conditions/return_code_is_not.sh
+++ b/minimizer/conditions/return_code_is_not.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
-cargo run -- --color never input.c 2>&1;
+source ./../run_cargo.sh
 ! [ $? = "$1" ]
--- a/minimizer/conditions/return_code_is_not.sh
+++ b/minimizer/conditions/return_code_is_not.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 # This is being executed inside of `rcc/minimizer/workspace/input.test` and thus requires `..` twice.
-source ./../../run_cargo.sh
+. ./../../run_cargo.sh
 ! [ $? = "$1" ]
--- a/minimizer/conditions/return_code_is_not.sh
+++ b/minimizer/conditions/return_code_is_not.sh
@@ -1,5 +1,3 @@
-#!/usr/bin/env bash
-
+#!/bin/sh
 cargo run -- --color never input.c 2>&1;
-if [[ $? != $1 ]]; then exit 0; fi
-exit 1;
+! [ $? = "$1" ]
--- a/minimizer/conditions/return_code_is_not.sh
+++ b/minimizer/conditions/return_code_is_not.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+cargo run -- --color never input.c 2>&1;
+if [[ $? != $1 ]]; then exit 0; fi
+exit 1;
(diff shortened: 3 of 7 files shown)


minimizer/conditions/output_contains.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cargo run -- --color never input.c 2>&1 | grep "$1"
+exit $?;
+
```
jyn514, line 5:
```suggestion
#!/bin/sh
cargo run -- --color never input.c 2>&1 | grep "$1"
```

minimizer/conditions/return_code_equals.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cargo run -- --color never input.c 2>&1;
+if [[ $? == $1 ]]; then exit 0; fi
+exit 1;
```
jyn514, line 5:
```suggestion
#!/bin/sh
cargo run -- --color never input.c 2>&1; 
[ $? = $1 ]
```

minimizer/conditions/return_code_is_not.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cargo run -- --color never input.c 2>&1;
+if [[ $? != $1 ]]; then exit 0; fi
+exit 1;
```
jyn514, line 5:
```suggestion
#!/bin/sh
cargo run -- --color never input.c 2>&1; 
! [ $? = $1 ]
```

minimizer/minimize.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+FILE=$1;
+CONDITION=$2;
```
jyn514, line 4:
```suggestion
FILE="$1"
CONDITION="$2"
```

minimizer/minimize.sh
```diff
@@ -0,0 +1,52 @@
...
+        filename=$(basename -- "$FILE")
+        extension="${filename##*.}"
+        filename="${filename%.*}"
+        filename="$SCRIPT_PATH/$filename-reduced.$extension";
+        
+        mv $WORKSPACE/input.reduced/input.c $filename;
+        echo "Ignore the above. The file is actually here: $filename";
+        rmdir $WORKSPACE/input.reduced;
+        rmdir $WORKSPACE;
+        exit 0;
```
jyn514, line 48:
```suggestion
```

Byter09:
> I haven't played around with this yet, I recommend running `shellcheck` on all these scripts though. Also, is it possible to relax the bound on `minimize.sh` so that it doesn't require the 2nd and 3rd args to be quoted?

I'm essentially looking for something like a variadic argument in bash. `$@` looks like it could do the job. Am investigating it while resolving the above comments.

jyn514:
`"$@"` sounds right, make sure you quote it.

Byter09:
It was `$*` (shellcheck helped there!)

Byter09:
I keep finding small errors. Hope you don't mind me not putting these in extra commits.

minimizer/conditions/output_contains.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+cargo run -- --color never input.c 2>&1 | grep "$1"
```
jyn514, line 2:
```suggestion
cargo run --quiet -- --color never --max-errors 0 input.c 2>&1 | grep "$1"
```

minimizer/conditions/return_code_equals.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+cargo run -- --color never input.c 2>&1;
```
jyn514, line 2:
```suggestion
cargo run --quiet -- --color never --max-errors 0 input.c 2>&1
```

minimizer/conditions/return_code_is_not.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+cargo run -- --color never input.c 2>&1;
```
jyn514, line 2:
```suggestion
cargo run --quiet -- --color never --max-errors 0 input.c 2>&1
```

minimizer/minimize.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+FILE="$1";
+CONDITION_SCRIPT="$2";
+shift;
+shift;
+CONDITION="$*";
```
jyn514, line 7: I'm not a giant fan of semicolons at the end of lines in bash, could you remove them in the rest of the program as well? If you keep it this way then either its more work the next time you modify it or it will be inconsistent if someone else doesn't add semicolons.
```suggestion
CONDITION_SCRIPT="$2"
shift
shift
CONDITION="$*"
```

minimizer/minimize.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
```
jyn514, line 2:
```suggestion
if [ $# -lt 2 ]; then
    echo "usage: $0 <file> <condition>"
    echo "note: see 'minimizer/conditions' for valid conditions"
    exit 1
fi

```

It would also be nice to move this check later so it can point to `$SCRIPT_PATH/conditions` instead, but that's a little to much to do from the github editor :laughing:

minimizer/minimize.sh
```diff
@@ -0,0 +1,52 @@
...
+    rmdir "$WORKSPACE";
+
+    exit 1;
+else
+    if ! [[ -f "$WORKSPACE/input.reduced/input.c" ]]; then
+        echo "DustMite failed. Cleaning up...";
+        rmdir "$WORKSPACE"/input.reduced;
+        rmdir "$WORKSPACE";
+        exit 1;
+    else
```
jyn514, line 39: Can you combine this into one check that does `rm -r $WORKSPACE` if it failed?

minimizer/minimize.sh
```diff
@@ -0,0 +1,52 @@
...
+fi
+
+if ! test -f "$FILE"; then
+    echo "Input file '$FILE' not found.";
+    exit 1;
+fi
+
+WORKSPACE=$SCRIPT_PATH/workspace
+mkdir "$WORKSPACE";
+cp "$FILE" "$WORKSPACE"/input.c;
```
jyn514, line 23: Maybe make this a symbolic link to reduce IO?
```suggestion
ln -s "$FILE" "$WORKSPACE"/input.c;
```

minimizer/minimize.sh
```diff
@@ -0,0 +1,52 @@
...
+
+        filename=$(basename -- "$FILE")
+        extension="${filename##*.}"
+        filename="${filename%.*}"
+        filename="$SCRIPT_PATH/$filename-reduced.$extension";
+
+        mv "$WORKSPACE"/input.reduced/input.c "$filename";
+        echo "Ignore the above. The file is actually here: $filename";
+        rmdir "$WORKSPACE"/input.reduced;
+        rmdir "$WORKSPACE";
```
jyn514, line 50:
```suggestion
        rm -r "$WORKSPACE"
```

jyn514:
You may also want to add a helper script with `cargo run --quiet -- --color never --max-errors 0` so you don't have to retype it for every condition.

minimizer/minimize.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
```
jyn514, line 1:
```suggestion
#!/usr/bin/env bash

set -euo pipefail
```
http://redsymbol.net/articles/unofficial-bash-strict-mode/

Byter09:
> You may also want to add a helper script with `cargo run --quiet -- --color never --max-errors 0` so you don't have to retype it for every condition.

Not quite sure this is a good idea as it would either be in the conditions folder but not actually be a condition, or it would be next to `minimizer.sh`, further polluting the nice and clean setup. Debatable though, depends on where you want to put that helper.

minimizer/minimize.sh
Byter09: I'm adding them to stay close to other languages and as a safety guard that no accidental line-fusing can happen. Will do it though if its "idiomatic".

jyn514: > no accidental line-fusing can happen

Newlines in bash always count as a statement terminator (unlike JavaScript or Haskell, which have automatic semicolon insertion).

> Will do it though if its "idiomatic"

Yes, it looks pretty odd to me this way.

jyn514:
> it would be next to `minimizer.sh`, further polluting the nice and clean setup

I think this is fine, there's only 3 files there currently and we could make the script non-executable so it's even more clear. The main reason I want it is to reduce the maintenance burden if we ever change the way we run `rcc`.

minimizer/minimize.sh
Byter09: I changed a bit more and also the ordering. do you mind checking this and resolving this if its okay?

minimizer/minimize.sh
```diff
@@ -0,0 +1,73 @@
...
+fi
+
+shift
+shift
+CONDITION="$*"
+WORKSPACE=$SCRIPT_PATH/workspace
+mkdir "$WORKSPACE"
+ln -s "$(realpath "$FILE")" "$WORKSPACE"/input.c
+
+dustmite --strip-comments --split "*.c:d" "$WORKSPACE"/input.c "$SCRIPT_PATH/conditions/$CONDITION_SCRIPT.sh $CONDITION"
```
jyn514, line 48: Hmm ... this doesn't work if I quote the condition:
```
./minimize.sh ../crash.c output_contains 'stack overflow'
+ dustmite --strip-comments --split '*.c:d' /home/joshua/src/rust/rcc/minimizer/workspace/input.c '/home/joshua/src/rust/rcc/minimizer/conditions/output_contains.sh stack overflow'
Loading /home/joshua/src/rust/rcc/minimizer/workspace/input.c
None => No
object.Exception@dustmite.d(270): Initial test fails (try using --no-redirect for details)
```
I tried `"$@"`, `"$*"`, and `$*` and all of them failed one way or another. The issue is that dustmite requires the command to be a single argument instead of multiple, normally `"$@"` would work for this but it creates multiple arguments instead of one. The awful hack is

```suggestion
dustmite --strip-comments --split "*.c:d" "$WORKSPACE"/input.c "$(printf " %q" "$SCRIPT_PATH/conditions/$CONDITION_SCRIPT.sh" "$@")"
```

This preserves the quoting behavior even if the original string passed by the user had spaces or quotes. See https://stackoverflow.com/questions/10835933/preserve-quotes-in-bash-arguments.

minimizer/minimize.sh
jyn514: Looks great :) I love how it prints the available conditions.

minimizer/minimize.sh
```diff
@@ -0,0 +1,73 @@
...
+if ! test -f "$FILE"; then
+    echo "Input file '$FILE' not found."
+    exit 1
+fi
+
+SCRIPT_PATH=$(cd "$(dirname "${BASH_SOURCE[0]}")" || exit; pwd -P)
+CONDITION_SCRIPT="$2"
+
+if ! test -f "$SCRIPT_PATH/conditions/$CONDITION_SCRIPT.sh"; then
+    echo "Condition script '$SCRIPT_PATH/conditions/$CONDITION_SCRIPT.sh' not found."
```
jyn514, line 35: You use `$SCRIPT_PATH/conditions/$CONDITION_SCRIPT.sh`, would it be easier to do `CONDITION_SCRIPT="$SCRIPT_PATH/conditions/$2.sh"` instead?

minimizer/minimize.sh
```diff
@@ -0,0 +1,73 @@
...
+if ! test -f "$CONDITION_SCRIPT"; then
+    echo "Condition script '$CONDITION_SCRIPT' not found."
+    echo "available conditions are:"
+    printf '\t%s\n' "${available_scripts[@]}"
+    exit 1
+fi
+
+shift
+shift
+CONDITION="$*"
```
jyn514, line 43:
```suggestion
```
This is no longer used

PR #407, opened 2020-05-06
Pretty Printing 2: Electric Boogaloo

pythondude325:
Fixed pretty printing so that it actually works. Also added anonymous struct & union printing, although with bad wight-space (it needs another PR to fix). I haven't added this yet, but I want to implement some round trip testing for the system.

### DIFF
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -1374,12 +1374,13 @@ pub(crate) mod test {
         analyze(s, Parser::expr, Analyzer::parse_expr)
     }
 
-    fn assert_decl_display(left: &str, right: &str) {
+    pub(crate) fn assert_decl_display(left: &str, right: &str) {
         assert_eq!(decl(left).unwrap().to_string(), right);
     }
     fn assert_extern_decl_display(s: &str) {
         // TODO: this `auto` is such a hack
-        assert_decl_display(s, &format!("auto {}", s));
+        // TOOD: We really should make this work.
+        assert_decl_display(s, &format!("{}", s));
     }
 
     pub(super) fn assert_same(left: &str, right: &str) {
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -1379,7 +1379,6 @@ pub(crate) mod test {
     }
     fn assert_extern_decl_display(s: &str) {
         // TODO: this `auto` is such a hack
-        // TOOD: We really should make this work.
         assert_decl_display(s, &format!("{}", s));
     }
 
--- a/src/data/hir.rs
+++ b/src/data/hir.rs
@@ -476,7 +476,9 @@ impl Display for Metadata {
         if self.qualifiers != Qualifiers::default() {
             write!(f, "{} ", self.qualifiers)?;
         }
-        write!(f, "{} ", self.storage_class)?;
+        if self.storage_class != StorageClass::default() {
+            write!(f, "{} ", self.storage_class)?;
+        }
         super::types::print_type(&self.ctype, Some(self.id), f)
     }
 }
--- a/src/data/mod.rs
+++ b/src/data/mod.rs
@@ -95,9 +95,9 @@ mod tests {
         let types = [
             "int",
             "int *",
-            "int[1][2][3]",
-            "char * (*)(float)",
-            "short * (*)[1][2][3]",
+            "int [1][2][3]",
+            "char *(*)(float)",
+            "short *(*)[1][2][3]",
             "_Bool",
             "struct s",
         ];
--- a/src/data/types.rs
+++ b/src/data/types.rs
@@ -281,17 +281,16 @@ pub(super) fn print_type(
     name: Option<InternedStr>,
     f: &mut Formatter,
 ) -> fmt::Result {
-
     print_declarator(ctype, name, f)?;
 
-    return Ok(());
+    Ok(())
 }
 
 fn write_struct_type(struct_type: &StructType, f: &mut Formatter) -> fmt::Result {
     match struct_type {
         StructType::Named(name, _) => {
             write!(f, "{}", name)?;
-        },
+        }
         StructType::Anonymous(members) => {
             write!(f, "{{ ")?;
             for member in members.iter() {
@@ -301,12 +300,10 @@ fn write_struct_type(struct_type: &StructType, f: &mut Formatter) -> fmt::Result
         }
     }
     Ok(())
-} 
+}
 
 pub fn print_declarator(ctype: &Type, name: Option<InternedStr>, f: &mut Formatter) -> fmt::Result {
-    fn unroll_type(
-        ctype: &Type,
-    ) -> Vec<&Type> {
+    fn unroll_type(ctype: &Type) -> Vec<&Type> {
         let mut types = Vec::new();
         let mut next_type = ctype;
         loop {
@@ -314,17 +311,15 @@ pub fn print_declarator(ctype: &Type, name: Option<InternedStr>, f: &mut Formatt
             next_type = match next_type {
                 Type::Array(of, _) => of.as_ref(),
                 Type::Pointer(to, _) => to.as_ref(),
-                Type::Function(FunctionType {
-                    return_type, ..
-                }) => return_type.as_ref(),
+                Type::Function(FunctionType { return_type, .. }) => return_type.as_ref(),
                 _ => break,
             };
         }
-        return types;
+        types
     }
 
-    use Type::*;
     use std::fmt::Write;
+    use Type::*;
 
     let unrolled_type = unroll_type(ctype);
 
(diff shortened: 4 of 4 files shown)


src/analyze/mod.rs
```diff
@@ -1374,12 +1374,13 @@ pub(crate) mod test {
...
     }
 
-    fn assert_decl_display(left: &str, right: &str) {
+    pub(crate) fn assert_decl_display(left: &str, right: &str) {
         assert_eq!(decl(left).unwrap().to_string(), right);
     }
     fn assert_extern_decl_display(s: &str) {
         // TODO: this `auto` is such a hack
-        assert_decl_display(s, &format!("auto {}", s));
+        // TOOD: We really should make this work.
```
jyn514, line 1382: I know this is bad, but I think it only needs one comment :laughing: 
```suggestion
```

src/data/hir.rs
```diff
@@ -476,7 +476,9 @@ impl Display for Metadata {
         if self.qualifiers != Qualifiers::default() {
             write!(f, "{} ", self.qualifiers)?;
         }
-        write!(f, "{} ", self.storage_class)?;
+        if self.storage_class != StorageClass::default() {
+            write!(f, "{} ", self.storage_class)?;
+        }
```
jyn514, line 481: This looks much better than I expected actually, I have no problem with this.

src/data/mod.rs
```diff
@@ -95,9 +95,9 @@ mod tests {
         let types = [
             "int",
             "int *",
-            "int[1][2][3]",
-            "char * (*)(float)",
-            "short * (*)[1][2][3]",
+            "int [1][2][3]",
+            "char *(*)(float)",
+            "short *(*)[1][2][3]",
```
jyn514, line 100: :tada:

src/data/types.rs
```diff
@@ -281,113 +281,165 @@ pub(super) fn print_type(
     name: Option<InternedStr>,
     f: &mut Formatter,
 ) -> fmt::Result {
-    print_pre(ctype, f)?;
-    print_mid(ctype, name, f)?;
-    print_post(ctype, f)
+    print_declarator(ctype, name, f)?;
+
+    Ok(())
 }
```
jyn514, line 287: I think we should get rid of this and rename `print_declarator` to `print_type`.

src/data/types.rs
```diff
@@ -281,113 +281,165 @@ pub(super) fn print_type(
...
+fn write_struct_type(struct_type: &StructType, f: &mut Formatter) -> fmt::Result {
+    match struct_type {
+        StructType::Named(name, _) => {
+            write!(f, "{}", name)?;
+        }
+        StructType::Anonymous(members) => {
+            write!(f, "{{ ")?;
+            for member in members.iter() {
+                write!(f, "{};", member)?;
+            }
```
jyn514, line 298: I know adding proper indentation requires a refactor, but can you add newlines?
```suggestion
            for member in members.iter() {
                writeln!(f, "    {};", member)?;
            }
```

src/data/types.rs
```diff
@@ -281,113 +281,165 @@ pub(super) fn print_type(
...
-        Type::Pointer(to, _) => print_post(to, f),
-        Type::Array(to, size) => {
-            write!(f, "[")?;
-            if let ArrayType::Fixed(size) = size {
-                write!(f, "{}", size)?;
-            }
-            write!(f, "]")?;
-            print_post(to, f)
+
+    let final_type = unrolled_type[unrolled_type.len() - 1];
```
jyn514, line 387: I think this would be nicer as `unrolled_type.pop().unwrap()` up at the top and then you iterate normally for the `enumerate` instead of skipping it. Alternatively you could return `(Vec<declarators>, specifier_type)` from `unroll_type`.

src/data/types.rs
```diff
@@ -281,113 +281,165 @@ pub(super) fn print_type(
...
 
-            write!(f, " ")?;
-            if depth {
-                write!(f, "(")?;
+    let mut prefixes = Vec::new();
+    let mut postfixes = Vec::new();
+
+    // Need to skip the last item because that's the final type that needs to be
+    // put in as the specifier
+    for (index, declarator_type) in unrolled_type[..unrolled_type.len() - 1].iter().enumerate() {
```
jyn514, line 331: Maybe this would be nicer using [`windows(2)`](https://doc.rust-lang.org/std/primitive.slice.html#method.windows)? That's mutually incompatible with my suggestion to use `pop()` though :laughing:

src/data/types.rs
```diff
@@ -281,113 +281,165 @@ pub(super) fn print_type(
...
+        Double => write!(f, "double")?,
+        Void => write!(f, "void")?,
+        Enum(Some(ident), _) => write!(f, "enum {}", ident)?,
+        Enum(None, _) => write!(f, "<anonymous enum>")?,
+        Union(struct_type) => {
+            write!(f, "union ")?;
+            write_struct_type(struct_type, f)?;
+        }
+        // Union(StructType::Named(ident, _)) => format!("union {}", ident),
+        // Union(_) => "<anonymous union>".to_string(),
```
jyn514, line 414:
```suggestion
```

src/data/types.rs
```diff
@@ -281,113 +281,165 @@ pub(super) fn print_type(
...
-            if func_type.varargs {
-                write!(f, ", ...")?;
-            }
-            write!(f, ")")
+        Bool => write!(f, "_Bool")?,
+        Float => write!(f, "float")?,
+        Double => write!(f, "double")?,
+        Void => write!(f, "void")?,
+        Enum(Some(ident), _) => write!(f, "enum {}", ident)?,
+        Enum(None, _) => write!(f, "<anonymous enum>")?,
```
jyn514, line 408: This would be a good follow-up PR :)

src/data/types.rs
```diff
@@ -281,113 +281,165 @@ pub(super) fn print_type(
...
+        }
+        // Union(StructType::Named(ident, _)) => format!("union {}", ident),
+        // Union(_) => "<anonymous union>".to_string(),
+        Struct(struct_type) => {
+            write!(f, "struct ")?;
+            write_struct_type(struct_type, f);
         }
-        _ => Ok(()),
+        // Struct(StructType::Named(ident, _)) => format!("struct {}", ident),
+        // Struct(_) => "<anonymous struct>".to_string(),
```
jyn514, line 420:
```suggestion
```

src/data/types.rs
```diff
@@ -281,113 +281,165 @@ pub(super) fn print_type(
...
+        VaList => write!(f, "va_list")?,
+        Error => write!(f, "<type error>")?,
+        _ => unreachable!(),
```
jyn514, line 423: Can you write out the declarators explicitly so it's more clear why they're unreachable? And maybe a comment as well.
```suggestion
        Array(_, _) | Function(_) | Pointer(_, _) => unreachable!("declarators were handled above"),
```

src/data/types.rs
pythondude325: Forgot to do that

src/data/types.rs
pythondude325: Sure.

src/data/types.rs
pythondude325: I was going to change it, but the way I detect whether or not parentheses are needed depends on the vector holding the complete type. I might still change it, but it will require more code than the currently very elegant parenthesis detection.

src/data/types.rs
pythondude325: I also considered windows, you can't use windows with a for loop, so I figure this would be nicer.

src/data/types.rs
jyn514: Hmm I see what you mean, it would have to be `unrolled_type.get(i + 1).unwrap_or(specifier_type)`.

src/data/types.rs
jyn514: I don't see why not? `Windows` is also an iterator: https://doc.rust-lang.org/std/slice/struct.Windows.html#impl-ExactSizeIterator

src/data/types.rs
pythondude325: Should I be using four spaces or a tab?

jyn514: Four spaces, same as rust code. Configuring whitespace is a whole different issue haha

src/data/types.rs
```diff
@@ -481,7 +481,7 @@ pub(crate) mod tests {
 
     #[test]
     fn test_big_one() {
-        assert_decl_display("struct { int i; } S;", "struct { int i; } S;");
+        assert_decl_display("struct { int i; } S;", "struct {    int i;\n} S;");
```
jyn514: This is still not correct, it will look like this when printed:
```c
struct {    int i;
} S;
```
Instead it should be
```c
struct {
    int i;
} S;
```

src/data/types.rs
```diff
@@ -491,6 +491,7 @@ pub(crate) mod tests {
         assert_decl_display(
             "int *(*jynelson)(int (*)(int));",
             "int *(*jynelson)(int (*)(int));",
-        )
+        );
+        assert_decl_display("int f(...);", "int f(...);");
```
jyn514, line 495: This should be caught as an error, so it will need to be removed when I fix that bug. It's fine for now though.

jyn514: Opened https://github.com/jyn514/rcc/issues/410

PR #414, opened 2020-05-07
Disable cranelift optimizations to avoid bugs

### LINKED ISSUE #404: [ICE] codegen error for while in switch
### Code

This is an upstream bug in cranelift: https://github.com/bytecodealliance/wasmtime/issues/1648

```c
void foo(void) {
  switch (1) {
    while(1) case 1:;
    case 0:;
  }
}
Message:  assertion failed: !self.opcode().is_branch()
Location: /home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/cranelift-codegen-0.63.0/src/ir/instructions.rs:254
```

### Expected behavior

Should loop forever at runtime.

jyn514:
r? @pythondude325 

Closes #404

### DIFF
--- a/src/ir/mod.rs
+++ b/src/ir/mod.rs
@@ -51,10 +51,6 @@ pub(crate) fn get_isa(jit: bool) -> Box<dyn TargetIsa + 'static> {
     flags_builder
         .enable("enable_verifier")
         .expect("enable_verifier should be a valid option");
-    // minimal optimizations
-    flags_builder
-        .set("opt_level", "speed")
-        .expect("opt_level: speed should be a valid option");
     // don't emit call to __cranelift_probestack
     flags_builder
         .set("enable_probestack", "false")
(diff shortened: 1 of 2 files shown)


jyn514:
I actually thought they were disabled already, it turns out that `speed` means the speed of the generated binary, not compilation speed.

src/ir/mod.rs
```diff
@@ -51,10 +51,6 @@ pub(crate) fn get_isa(jit: bool) -> Box<dyn TargetIsa + 'static> {
     flags_builder
         .enable("enable_verifier")
         .expect("enable_verifier should be a valid option");
-    // minimal optimizations
-    flags_builder
-        .set("opt_level", "speed")
```
pythondude325, line 56: Should it be set to "none" instead, or is none a well defined default?

jyn514: 'well defined' is a little unclear in the context of cranelift ... it's just as likely for `none` to be removed as for the default to be changed in the next release. As long as we have the test suite I'll notice so I think it's ok. See also https://github.com/bytecodealliance/wasmtime/issues/1634

pythondude325: Cranelift releases are fun.

pythondude325:
I say you can go ahead and merge it.

PR #422, opened 2020-05-10
Use git-testament instead of rewriting it badly

jyn514:
This handles
- Changes to the repo that don't affect `build.rs`
- The repo not existing

It also shows a lot more info:
- Number of commits since last tag
- Number of modifications since last commit
- Date of last commit

It allows removing the `build.rs`, improving compile times.


jyn514:
Example output:
```
$ cargo run -q -- -V
rcc 0.8.0+310 (b1f0618f2 2020-05-10) dirty 27 modifications
```

PR #424, opened 2020-05-10
Ast declaration printing

pythondude325:
Fixed printing AST declarations

### DIFF
--- a/src/data/ast.rs
+++ b/src/data/ast.rs
@@ -669,3 +669,14 @@ impl Display for Expr {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::parse::decl::test::assert_no_change;
+
+    #[test]
+    fn test_declaration_display() {
+        assert_no_change("int (*(*f))();");
+    }
+}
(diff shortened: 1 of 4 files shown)


src/data/ast.rs
```diff
@@ -436,78 +436,64 @@ impl Declarator {
...
-                write!(f, "({}", joined(params, ", "))?;
-                if *varargs {
-                    write!(f, ", ...")?;
+                DeclaratorType::Function(function_definition) => {
+                    write!(f, "({}", joined(function_definition.params.iter(), ", "))?;
+                    if function_definition.varargs {
+                        write!(f, ", ...")?;
+                    }
+                    write!(f, ")")?;
+                }
```
jyn514, line 488: This is a declarator, not a definition. A declarator is `f(int)`, a definition is `int f(int a) { return 1; }`. All definitions are also declarations, but not all declarations are definitions, e.g. `int f(int);`.

src/data/hir.rs
```diff
@@ -528,7 +528,10 @@ impl Eq for Symbol {}
 
 #[cfg(test)]
 mod tests {
-    use crate::parse::test::parser;
+    use crate::analyze::test::analyze;
+    use crate::Analyzer;
+    use crate::Locatable;
+    use crate::Parser;
```
jyn514, line 534:
```suggestion
    use crate::{Analyzer, Locatable, Parser};
```

PR #431, opened 2020-05-13
Make the top-level API much easier to use

jyn514:
- Return `Files` instead of requiring it to be provided
- Use `Program` struct instead of an tuple and type alias
- Allow `filename` to be passed so that local includes work properly
     - This also moves `filename` from `BinOpt` to `Opt`

### DIFF
--- a/src/data/lex.rs
+++ b/src/data/lex.rs
@@ -427,9 +427,7 @@ pub(crate) mod test {
     }
     /// Create a new preprocessor with `s` as the input, but without a trailing newline
     pub(crate) fn cpp_no_newline(s: &str) -> PreProcessor {
-        let mut files: Files = Default::default();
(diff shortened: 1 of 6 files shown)


jyn514:
For context, this is because I was reading through https://github.com/rust-lang/rust-bindgen/blob/d65100ebbaffabcc91e81ba3e88160de7cb2d5ff/src/ir/var.rs#L323 for https://github.com/rust-lang/rust-bindgen/issues/1762 and noticed that the cexpr API is _much_ nicer than what we had before. Now ours is almost as nice:
```rust
let cpp = rcc::PreProcessorBuilder::new("1 + 2").build();
// TODO: make cpp_expr() public
let expr = cpp.cpp_expr()?;
```

Issue #395, opened 2020-05-01
\n is a valid preprocessing token


jyn514:
### Expected behavior

`\n` should be a valid preprocessor token so that it can be used in token pasting.
Found in `/usr/include/sys/sysmacros.h`. glibc is cursed.

### Code

```c
#define __SYSMACROS_DM(symbol) __SYSMACROS_DM1 (a b \n c)
#define __SYSMACROS_DM1(...) __glibc_macro_warning (#__VA_ARGS__)
__SYSMACROS_DM(a)
<stdin>:1:53 error: invalid token: unknown token: '\'
#define __SYSMACROS_DM(symbol) __SYSMACROS_DM1 (a b \n c)
                                                    ^
```

Output from `clang -E`:
```c
__glibc_macro_warning ("a b \n c")
```

http://port70.net/~nsz/c/c11/n1570.html#6.4p3:
> The categories of preprocessing tokens are: header names, identifiers, preprocessing numbers, character constants, string literals, punctuators, and single non-white-space characters that do not lexically match the other preprocessing token categories.

"If it's not whitespace, it's valid!" :facepalm:

jyn514:
Fixing this long-term will be difficult, since `\n` is _not_ a valid lexer token. I came up with an idea to split apart the lexer and preprocessor: preprocessor tokens are a superset of lexer tokens, but should be turned _into_ lexer tokens by the time preprocessing finishes. Then there would be two options for a lexer: either the preprocessor or the lexer, but not both.

hdamron17:
So is '

Issue #361, opened 2020-03-30
[ICE] unput doesn't play well with consume_whitespace


jyn514:
### Code

Note the error message is from a local copy, I plan to make a PR soon.

```c
#i ""
/b
The application panicked (crashed).
Message:  unputting '\n' would cause the lexer to forget it saw 'b' (current is '/')
Location: src/lex/mod.rs:153
```

### Expected behavior

The lexer should output the tokens `Hash`, `Id("i")`, `Str("")`, `Slash`, and `Id("b")`. Additionally, `seen_line_token` should be set appropriately at all times.

The reason this is hard is because I only want to peek ahead 2 characters, but at the same time the preprocessor needs to know where newlines occur. I think the real fix will be to implement `\n` as a token (#356).

Extended description (from discord):

The way the lexer works is it's streaming, it looks at one byte at a time.
Sometimes it needs to look at multiple, but it doesn't have a buffer to store them in, so instead it uses `current` for one byte ahead and `lookahead` for 2 bytes ahead.
The culprit is `consume_whitespace`, it calls `peek_next` when it sees `/` which sets `lookahead`.
The preprocessor cares about newlines, even though the lexer doesn't,
so the lexer still needs to keep track of '\n', which involves setting `seen_line_token`.
As a result `parse_string` has to unput a newline after `consume_whitespace` gets rid of it in order for `seen_line_token` to get appropriately.

See https://github.com/jyn514/rcc/blob/master/src/lex/mod.rs#L610 for more details.

<details><summary>Backtrace</summary>

```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                          (5 post panic frames hidden)                          
 5: rcc::lex::Lexer::unput::h70bde51f6a37bfb0
    at src/lex/mod.rs:153
 6: rcc::lex::Lexer::parse_string::hce2d12999d596f68
    at src/lex/mod.rs:619
 7: <rcc::lex::Lexer as core::iter::traits::iterator::Iterator>::next::{{closure}}::hb2227ef3fa992171
    at src/lex/mod.rs:845
 8: core::option::Option<T>::and_then::hce18bbfb15e478ce
    at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447/src/libcore/option.rs:658
[...]

PR #437, opened 2020-05-15
Close #356 - Add whitespace token for preprocessor

### LINKED ISSUE #356: Remember whitespace for -E
This would require adding a whitespace token to the lexer. This has a few advantages:

- The output from `-E` would be much more readable
- `-E` could be abused for non-C-like tokens (cc @Nemo157)
- The preprocessor could be decoupled from the lexer, helping with #266 (cc @pythondude325)

hdamron17:
When complete, this will close #356 and maybe #395.

So far, the whitespace token has been added and properly outputs when using `-E`, but I still need to consume the whitespace before it's used by the parser.


jyn514:
I'd be interested if it helped with #361 as well.

hdamron17:
Observation: Need to add whitespace consideration back to preprocessor (e.g. `#ifndef x` fails because of the whitespace token in the middle) but ensure whitespace tokens have no newlines. Unit tests also fail a lot because the comparisons do not take whitespace into account.

src/lex/mod.rs
```diff
@@ -211,6 +214,7 @@ impl Lexer {
                 break;
```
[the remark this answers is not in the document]
hdamron17: That's what I would have thought too, but `clang -E` does not preserve comments. Not sure about `gcc` since I'm on a Mac.

src/main.rs
```diff
@@ -134,7 +134,7 @@ fn real_main(buf: Rc<str>, bin_opt: BinOpt, output: &Path) -> Result<(), (Error,
         let stdout = io::stdout();
         let mut stdout_buf = BufWriter::new(stdout.lock());
         for token in rcc_try!(tokens, files) {
-            write!(stdout_buf, "{} ", token.data).expect("failed to write to stdout");
+            write!(stdout_buf, "{}", token.data).expect("failed to write to stdout");
         }
         writeln!(stdout_buf).expect("failed to write to stdout");
```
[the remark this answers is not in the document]
hdamron17: Presumably the original C file ended with a newline so this would maintain the ending newline. What does this change do?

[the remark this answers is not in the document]
jyn514: Before the output for `-E` looked like this:
```
int main() {
	return 1;
}

```
Now it will look like this:
```
int main() {
	return 1;
}
```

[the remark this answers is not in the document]
jyn514: In other words, originally you would print both the original ending newline and add a newline of your own.

src/lex/mod.rs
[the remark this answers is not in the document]
jyn514: Hmm apparently `gcc -E` also strips comments. Ok, I guess this is fine then.

jyn514:
Note this will not fix https://github.com/jyn514/rcc/issues/395. The `\n` in the issue title refers to the actual characters `\` and then `n`, not a newline.

hdamron17:
It's working now, but I probably should put some whitespace-dependent test cases for `-E`.

One possible issue that came up is that the parser iteration does not work if there is leading whitespace. However, it works in the final product so I don't think it matters. I changed the test case to have no leading whitespace in 8e0509e.

hdamron17:
Yeah it doesn't keep whitespace properly, mostly with preprocessor stuff since I was just trying to get existing tests to pass...

jyn514:
> One possible issue that came up is that the parser iteration does not work if there is leading whitespace. However, it works in the final product so I don't think it matters. I changed the test case to have no leading whitespace in 8e0509e.

If I understand right, this means that it works correctly using `check_semantics` but that passing a whitespace token to `Parser` as `first` will give a spurious error? That seems fine as long as it's documented.

src/lex/cpp.rs
```diff
@@ -886,6 +886,15 @@ impl<'a> PreProcessor<'a> {
...
                 .map(|res| res.map(|loc| loc.data))
+                .enumerate()
+                .map(|(i, x)| {
+                    if i == 0 {
+                        vec![x]
+                    } else {
+                        vec![x, Ok(Token::Whitespace(String::from(" ")))]
+                    }
+                })
+                .flatten()
```
jyn514, line 897: Why do you add these spaces here in the define? Does `tokens_until_newline` not return whitespace tokens?

hdamron17: `tokens_until_newline` (at the moment) does not include whitespace tokens. I did not want to change it since it is somewhat separate from the preprocessor. I can try adding whitespace tokens to it and see what happens now that it is in a stable state. I did notice that `clang` only puts one space when replacing preprocessor defines, regardless of the original spacing.

jyn514: Hmm interesting ... I suppose since the behavior is correct we can try to go back and improve the spacing later.

jyn514: Also, please add a comment to this effect either here or at `tokens_until_newline`.

src/lex/mod.rs
```diff
@@ -191,22 +191,29 @@ impl Lexer {
...
-    fn consume_whitespace(&mut self) {
+    fn consume_whitespace(&mut self) -> String {
         // there may be comments following whitespace
+        let mut whitespace = String::new();
         loop {
             // whitespace
             while self.peek().map_or(false, |c| c.is_ascii_whitespace()) {
-                self.next_char();
+                if let Some(c) = self.next_char() {
+                    whitespace.push(c.into());
```
jyn514, line 201: This might now be better as a match + loop so you don't have to check the `Some` condition twice. Up to you though.

src/lex/mod.rs
```diff
@@ -865,7 +905,8 @@ impl Iterator for Lexer {
                         .with(LexError::UnknownToken(x as char))));
                 }
             };
-            self.seen_line_token |= data != Token::Hash;
+            // We've seen a token if this isn't # and this isn't whitespace
+            self.seen_line_token |= data != Token::Hash && !matches!(data, Token::Whitespace(_));
```
jyn514, line 909: This might be easier to read if you reversed the condition: `!(Token::Hash || whitespace)`.

src/lex/cpp.rs
hdamron17: I actually went back and did it the proper way by changing tokens_until_newline. Unfortunately I had to rework some stuff for boolean_expr because whitespace show up in the replacement stage. I'll push as soon as I double check the tests. Also, I think the rework will be much less of an eyesore.

src/lex/mod.rs
hdamron17: Not sure what you mean by that, but this doesn't look too bad to me.

src/lex/mod.rs
hdamron17: Done

hdamron17:
Some more tests with preprocessor stuff would be nice, but other than that, I think it's complete.

hdamron17:
Also, newlines are not preserved for preprocessor macros at the moment, so I'll need to fix that.

jyn514:
> Also, newlines are not preserved for preprocessor macros at the moment, so I'll need to fix that.

Do you mean that `\\\n` isn't preserved?
```c
#define f(a) { \
	int b = a; \
	return b; \
}
f(1)
 { int b = 1; return b; }
```
If so, that's fine, clang does the same. It will also be hard to do without major changes since deleting `\\\n` happens very early in the lexer: https://github.com/jyn514/rcc/blob/93b5e06/src/lex/mod.rs#L98

src/lex/cpp.rs
```diff
@@ -1215,19 +1256,34 @@ mod tests {
...
             "{} is not the same as {}",
             src,
             cpp_src,
         );
     }
+    fn assert_same_exact(src: &str, cpp_src: &str) {
+        // NOTE make sure `cpp_src` has a trailing newline
+        let pprint = cpp(src)
+            .filter_map(|res| res.ok().map(|token| token.data))
+            .map(|res| format!("{}", res))
```
jyn514, line 1282:
```suggestion
            .filter_map(|res| res.ok().map(|token| token.data.to_string()))
```

src/lex/cpp.rs
```diff
@@ -1534,4 +1589,12 @@ int main(){}
         ";
         assert_same(original, "a(1)");
     }
+    #[test]
+    // https://github.com/jyn514/rcc/issues/356
+    fn preprocess_only() {
+        assert_same_exact("int \t\n\r     main() {}", "int \t\n\r     main() {}");
+        assert_same_exact("int/* */main() {}", "int main() {}");
+        assert_same_exact("int/*\n\n\n*/main() {}", "int\n\n\nmain() {}");
```
jyn514, line 1597: Hmm, this behavior looks a little confusing. Is there a reason you kept newlines inside of block comments?

hdamron17:
> > One possible issue that came up is that the parser iteration does not work if there is leading whitespace. However, it works in the final product so I don't think it matters. I changed the test case to have no leading whitespace in [8e0509e](https://github.com/jyn514/rcc/commit/8e0509e45ef518c0c3b2583b92de0cc99aa0e096).
> 
> If I understand right, this means that it works correctly using `check_semantics` but that passing a whitespace token to `Parser` as `first` will give a spurious error? That seems fine as long as it's documented.

Turns out it was an easy fix. I just changed the `parser` function to use `next_non_whitespace`.

src/lex/cpp.rs
hdamron17: This seems to be the behavior of clang. For example,
```
int main() {
 /*




 */
}
```
preprocesses to
```
# 1 "test.c"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 363 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "test.c" 2
int main() {






}
```

hdamron17: I may need to look at the documentation though since
```
int main() {/*




*/}
```
mysteriously preprocesses to
```
# 1 "test.c"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 363 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "test.c" 2
int main() { }
```

hdamron17:
> > Also, newlines are not preserved for preprocessor macros at the moment, so I'll need to fix that.
> 
> Do you mean that `\\\n` isn't preserved?

No, I mean `#define`s do not keep newlines after.
```
int main() {
#define a
#define b
#define c
}
```
preprocesses to
```
int main() {
}
```
but should preprocess to
```
# 1 "test.c"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 363 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "test.c" 2
int main() {



}
```

hdamron17:
The newlines not being preserved after some macros is because `tokens_until_newline` does not actually stop at the newline but just consumes arbitrary whitespace until the line number increases. For example,
```
int main() {
#define a 1
           return a;
}
```
preprocesses to
```
int main() {
return 1;
}
```

jyn514:
> tokens_until_newline does not actually stop at the newline but just consumes arbitrary whitespace until the line number increases

I don't see a way to fix this without fully separating the preprocessor from the lexer ... I guess I could make that function part of the lexer instead and stop at `\n`? That would require a fair bit of refactoring but not an unreasonable amount.

jyn514:
In any case I think it can be fixed later, the majority of things are working now.

src/lex/cpp.rs
jyn514: Relevant part of the standard ([5.1.1.2 Translation phases](http://port70.net/~nsz/c/c11/n1570.html#5.1.1.2)):
>  The source file is decomposed into preprocessing tokens7) and sequences of white-space characters (including comments). A source file shall not end in a partial preprocessing token or in a partial comment. Each comment is replaced by one space character. New-line characters are retained. Whether each nonempty sequence of white-space characters other than new-line is retained or replaced by one space character is implementation-defined.

This seems to me to say that clang's behavior in your first example is correct, and the behavior in the second is a bug. For context, gcc always behaves like clang in the second example, i.e. it always deletes newlines in comments:

```c
gcc -x c -E -P -
int main() {
 /*




 */
}
```
preprocesses to
```c
int main() {
}
```
and
```c
int main() {/*




*/}
```
preprocesses to `int main() { }`.

jyn514: tcc behaves the same as gcc.

jyn514:
r=me once @hdamron17 is happy with it

hdamron17:
I went ahead and changed `tokens_until_newline` and it helps with `#defines`. All that's missing now are `#else` and `#endif` I think so I'll try to get those working and make enough tests to be worthwhile.

src/lex/cpp.rs
```diff
@@ -1594,6 +1595,8 @@ int main(){}
         assert_same_exact("int \t\n\r     main() {}", "int \t\n\r     main() {}");
         assert_same_exact("int/* */main() {}", "int main() {}");
         assert_same_exact("int/*\n\n\n*/main() {}", "int\n\n\nmain() {}");
-        // TODO add tests for `assert_same_exact` with preprocessor macros
+        assert_same_exact("#define a(c) c\tc\na(1);a(2)", "\n1\t1;2\t2");
+        // assert_same_exact("#define a //\n#if defined a\n  x\n#endif", "\n\n  x");
+        // TODO add mote tests for `assert_same_exact` with preprocessor macros
```
jyn514, line 1600:
```suggestion
        // TODO add more tests for `assert_same_exact` with preprocessor macros
```

src/lex/cpp.rs
```diff
@@ -880,17 +894,14 @@ impl<'a> PreProcessor<'a> {
     // `#define f (a) - object macro
     fn define(&mut self, start: u32) -> Result<(), Locatable<Error>> {
         let body = |this: &mut PreProcessor| {
-            this.tokens_until_newline()
+            this.tokens_until_newline(true)
                 .into_iter()
+                .skip(1)  // First is always unwanted whitespace
```
jyn514, line 899: Are you sure this is true? What about things like `#define f+`? This has different behavior on your branch than on master (and probably deserves an automated test).

src/lex/cpp.rs
```diff
@@ -1215,19 +1257,33 @@ mod tests {
             _ => panic!("not a keyword: {:?}", token),
         }
     }
+    fn _is_not_whitespace(res: &CompileResult<Token>) -> bool {
+        !matches!(res, Ok(Token::Whitespace(_)))
+    }
```
jyn514, line 1262: Why is there both this function and `is_not_whitespace`?

src/lex/cpp.rs
jyn514: Ah interesting, clang gives an error if there's not a space. We should do the same.

```
$ clang -x c - -E -pedantic-errors -P
#define f+
f
<stdin>:1:10: error: ISO C99 requires whitespace after the macro name [-Werror,-Wc99-extensions]
```

src/lex/cpp.rs
hdamron17: I was being dumb and applied the second one in a different order so I though I needed two with different signatures Changed in coming commit.

src/lex/cpp.rs
hdamron17: "warning: ISO C99 requires whitespace after the macro name"
I guess I should at least error here if it's not whitespace instead of blindly assuming it is.

hdamron17: Whitespace is no longer required between `#define` id and body. I do not give a warning if there is no whitespace, though.

hdamron17:
r @jyn514

hdamron17:
Your summary sounds about right. Also the behaviour of `consume_whitespace_oneline` was just copied from your code so I guess you're the clever one. :)

PR #438, opened 2020-05-18
Switch the string interner to lasso

Kixiron:
[`lasso`](https://github.com/Kixiron/lasso) is ~50% faster than `string-interner` and does not have the UB and Miri errors that `string-interner` currently does


jyn514:
Benchmarks: https://docs.rs/lasso/0.2.2/lasso/#benchmarks

PR #440, opened 2020-05-20
Separate macro replacement from macro parsing

### LINKED ISSUE #436: Separate macro replacement and macro parsing
Currently both replacement and parsing for the preprocessor are done at the same time, so you can't provide the preprocessor with a list of tokens and have those be replaced; instead you have to provide the raw string of the source code.

jyn514:
This allows you to replace any iterator of `Token` without having to have the original `&str`.

The core idea is to separate the preprocessor into 3 separate, nested iterators:
1. The innermost iterator (`FileProcessor`) deals with multiple files/lexers. If one included file runs out of tokens, it seamlessly goes on to the next one.
2. The middle iterator (`MacroReplacer`) performs macro replacement.
3. The outermost iterator (`PreProcessor`) deals with preprocessing directives. Additionally, because of the weird way I lumped the lexer and preprocessor together, it recognizes keywords and turns `Token::Id` into `Token::Keyword` as necessary.

The `PreProcessor` sometimes does _not_ want to replace its tokens (e.g. for `#if defined(a)`). In this case, it reaches  _through_ the replacer into the `FileProcessor` to drag out those tokens. Because of this, the `MacroReplacer` _cannot_ have any idea of pending tokens, or its `pending` will conflict with the `PreProcessor`'s  `pending`.

I want to support inputs other than `FileProcessor` (this is the whole point of the PR after all), but at the same time I need to be able to peek ahead in the input to see whether a token defined as a function macro is followed by a `(`. To support both these use cases, I add a new trait called `Peekable`; the intent is that users of `rcc` can use arbitrary iterators over `Token` by calling `iter.peekable()`.

I'm not super happy with how tightly coupled the `PreProcessor` is to the `FileProcessor`, but that's a project for another day.

This also implements `Debug` for `InternedStr` to actually be useful, since I was looking at the debug output a lot.

Needed for https://github.com/rust-lang/rust-bindgen/pull/1782. Closes #436.

r? @pythondude325

### DIFF
--- a/src/intern.rs
+++ b/src/intern.rs
@@ -4,8 +4,14 @@ use std::sync::RwLock;
 use lasso::{Rodeo, Spur};
 use lazy_static::lazy_static;
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
 pub struct InternedStr(pub Spur);
+
+impl fmt::Debug for InternedStr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self)
+    }
+}
 
 lazy_static! {
     pub static ref STRINGS: RwLock<Rodeo<Spur>> = RwLock::new(Rodeo::default());
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -40,11 +40,6 @@ use crate::data::*;
 use crate::get_str;
 use crate::Files;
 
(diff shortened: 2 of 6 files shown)


jyn514:
I realized as I wrote up the PR that most of the PR description should be doc-comments, adding those now.

jyn514:
I don't expect much of the frontend work to be target-dependent (other than `sizeof(int)` and that sort of thing). It's changes to the backend that are much more likely to break on different platforms, especially linking.

jyn514:
Closing in favor of https://github.com/jyn514/rcc/pull/441

PR #441, opened 2020-05-20
Make cpp_expr usable without having a full PreProcessor

jyn514:
Replaces #440


jyn514:
Changes from 440: `MacroReplacer` is no longer a thing and `replace` is now a free-standing function. This has the advantage that you no longer have to passed an owned `HashMap` of definitions, and also makes things better for the `PreProcessor` by allowing it to use `self.file_processor` instead of `self.replacer.inner` everywhere.

Issue #442, opened 2020-05-20
1.0/0 and 0.0/0 are both defined, but are reported as an error by rcc


jyn514:
### Expected behavior

`1.0 / 0` should be `+INF`, `0.0 / 0` should be `NaN`.

### Code

```c
<stdin>:1:11 error: invalid program: cannot divide by zero
float f = 1.0 / 0;
          ^^^^^^^
<stdin>:1:11 error: invalid program: cannot divide by zero
float f = 0.0 / 0;
          ^^^^^^^
```

Annex F: floating point arithmetic (http://port70.net/~nsz/c/c11/n1570.html#F)

> 1 The C floating types match the IEC 60559 formats as follows:
    - The float type matches the IEC 60559 single format.
    - The double type matches the IEC 60559 double format.
    - The long double type matches an IEC 60559 extended format,357) else a non-IEC 60559 extended format, else the IEC 60559 double format. 

We should also define `__STDC_IEC_559__`, as per [6.10.8.3](http://port70.net/~nsz/c/c11/n1570.html#6.10.8.3).

jyn514:
Found in https://github.com/rust-lang/rust-bindgen/pull/1782

jyn514:
Relevant code: https://github.com/jyn514/rcc/blob/24eaa840742a87c382f25ca9a0a4dba66affe549/src/fold.rs#L326

PR #444, opened 2020-05-21
Don't give an error for 1.0/0 and 0.0/0

jyn514:
Closes https://github.com/jyn514/rcc/issues/442

I also found https://github.com/rust-lang/rust/issues/72411 while working on this, so that was fun.

### DIFF
--- a/src/fold.rs
+++ b/src/fold.rs
@@ -323,7 +323,7 @@ fn fold_binary(
             Mul,
         ),
         Div => {
-            if right.is_zero() {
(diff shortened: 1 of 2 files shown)


jyn514:
I think a better solution would be to modify `store_static` to use a known bitpattern for NaN instead of blindly trusting `to_bits`.

jyn514:
Delaying the above ^ until someone complains, since it's not defined by any standard.

PR #445, opened 2020-05-22
Document InternedStr

jyn514:
r? @Kixiron

### DIFF
--- a/src/intern.rs
+++ b/src/intern.rs
@@ -6,6 +6,11 @@ use lazy_static::lazy_static;
 
 /// A opaque identifier for a string which has been [interned].
 ///
+/// Interning strings means they are cheap to copy and compare,
+/// at the cost of taking an O(n) hash comparison to intern.
+/// Interning also reduces memory usage for programs
+/// with many identifiers which are repeated often (i.e. C headers).
+///
 /// [interned]: https://en.wikipedia.org/wiki/String_interning
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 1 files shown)


src/intern.rs
```diff
@@ -4,6 +4,9 @@ use std::sync::RwLock;
 use lasso::{Rodeo, Spur};
 use lazy_static::lazy_static;
 
+/// A opaque identifier for a string which has been [interned].
```
Kixiron, line 7: Maybe go into a little more detail on _why_ strings are being interned, someone not familiar may not understand the reasoning at a glance

src/intern.rs
```diff
@@ -24,12 +39,28 @@ macro_rules! get_str {
...
+    /// Convert this identifier back into the original `String`, cloning it along the way.
+    ///
+    /// # See also
+    /// [`get_str!`](../macro.get_str.html)
+    ///
+    /// # Panics
+    /// This function will panic if another thread panicked while accessing the global string pool.
+    ///
     pub fn resolve_and_clone(self) -> String {
         get_str!(self).to_string()
```
Kixiron, line 55: Not necessarily part of the change, but why is `.to_string()` used when they function is about cloning? They’re functionally the same, but it seems odd

jyn514: `get_str!` returns a `&str`, not a `String`.

Kixiron: Ah, indeed it does, I always forget that clone on `&str` doesn’t do anything

PR #448, opened 2020-05-23
Make git-testament optional and disable it by default

jyn514:
This allows compiling rcc on alpine, which doesn't support proc-macros.
git-testament is mostly useful for debug builds anyways.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ proptest-derive = "0.1"
 [features]
 default = ["cc", "codegen", "color-backtrace"]
 # The `swcc` binary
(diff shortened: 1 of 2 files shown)


jyn514:
I just found that this doesn't help because `thiserror-impl` is a proc-macro, and we have a hard dependency on thiserror. This makes things tricky.

jyn514:
Status: https://github.com/rust-lang/rust/issues/40174#issuecomment-632957034

jyn514:
Going to merge this anyway since I think it's good to reduce compile times. I wish cargo had dev-dependencies for `lib.rs` :/

Issue #384, opened 2020-04-27
The lexer should only return a span and a token-kind


jyn514:
This is the traditional way to write a lexer, used by `yacc`, `logos`, and `clang`. The lexer identifies the boundaries of the token and the token kind and _nothing else_. This means for instance `1000000000000000000000000` would not be a lex error, but instead a parse error (and I'd probably want to delay it even further into constant folding).

This would make it significantly easier to switch to logos and support https://github.com/rust-lang/rust-bindgen/issues/1762.

This requires a fairly significant refactor.

hdamron17:
Why should the lexer only provide span and token kind? It seems this restriction is only necessary for the preprocessor.

jyn514:
Look at the linked issues for examples; #385 and #383 in particular would be tricky without this, I'd have to move types all the way into the lexer.

Stupremee:
After I finished my current PRs, I will try to implement this.

hdamron17:
I am working on this in #474 because I needed it for #456.

PR #456, opened 2020-05-25
Preprocessor stringify #

hdamron17:
This is an extension of branch in #437. I will rebase when that branch is merged.

Initial attempt seems to work for all cases except stringification of strings which contain escaped characters, in which case the escape is not deep enough. E.g.
```
#define f(a) #a
f("\n")
```
preprocesses to
```

"\"\n\""
```
but should be
```

"\"\\n\""
```

### DIFF
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -857,6 +857,10 @@ impl<'a> PreProcessor<'a> {
                     location,
                 ),
             }
+            self.consume_whitespace_oneline(
+                self.file_processor.offset(),
+                CppError::Expected("',' or ')'", "macro parameter list"),
+            )?;
             // either `,` or `)`
             if self.lexer_mut().match_next(')') {
                 return Ok(arguments);
@@ -1798,5 +1802,13 @@ h",
         assert_same_stringified(r#"'\n'"#, r#""'\\n'""#);
         assert_same_stringified(r#"  a +   b"#, r#""a + b""#);
         assert_same_stringified("", r#""""#);
+        assert_same_exact(
+            r#"#define f(  x  ,y  )   4 # x ; #y
+          f (   42 ,  "hey there world" ) + f(1,0) "#,
+            r#"
+          4 "42" ; "\"hey there world\"" + 4 "1" ; "0" "#,
+        );
+        assert!(cpp("#define f(x) #y\nf(0)").any(|x| x.is_err()));
+        assert!(cpp("#define f(x) #+\nf(0)").any(|x| x.is_err()));
     }
 }
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -1770,22 +1770,31 @@ h",
             .join("");
         assert_eq!(cpp, "\nf");
     }
+    fn assert_same_stringified(cpp: &str, cpp_src: &str) {
+        assert_same_exact(
+            &format!("#define xstr(a) #a\nxstr({})", cpp),
+            &format!("\n{}", cpp_src),
+        );
+    }
     #[test]
-    fn preprocess_stringify() {
-        let assert_same_stringified = |cpp: &str, cpp_src: &str| {
-            assert_same_exact(
-                &format!("#define xstr(a) #a\nxstr({})", cpp),
-                &format!("\n{}", cpp_src),
-            );
-        };
+    fn stringify() {
         assert_same_stringified("a + b", r#""a + b""#);
         assert_same_stringified(r#"b   	 +"c""#, r#""b +\"c\"""#);
         assert_same_stringified(r#""+\\+\n+\"+""#, r#""\"+\\\\+\\n+\\\"+\"""#);
-        assert_same_stringified(r#""\'""#, r#""\"\\'\""#);
+        assert_same_exact("#define xstr(a, b) #a = b\nxstr(1+2,3)", "\n\"1+2\" = 3");
+        assert_same_exact("#define xstr(a, b) a b\nxstr(1+2,3)", "\n1+2 3");
+        assert_same_exact("#define xstr(a) # a\nxstr(1+2)", "\n\"1+2\"");
+        assert_same_exact("#define hash #a\nhash", "\n#a");
+    }
+    #[test]
+    #[ignore = "Depends on #384"]
+    fn stringify_string() {
+        assert_same_stringified(r#""\'""#, r#""\"\\'\"""#);
         assert_same_stringified(
             r#""\a+\b+\e+\f+\r+\v+\?""#,
             r#""\"\\a+\\b+\\e+\\f+\\r+\\v+\\?\"""#,
         );
         assert_same_stringified(r#""\x3f""#, r#""\"\\x3f\"""#);
+        assert_same_stringified(r#""\xff""#, r#""\"\\xff\"""#);
     }
 }
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -1770,4 +1770,17 @@ h",
             .join("");
         assert_eq!(cpp, "\nf");
     }
+    #[test]
+    fn preprocess_stringify() {
+        let assert_same_stringified = |cpp: &str, cpp_src: &str| {
+            assert_same_exact(
+                &format!("#define xstr(a) #a\nxstr({})", cpp),
+                &format!("\n{}", cpp_src));
+        };
+        assert_same_stringified("a + b", r#""a + b""#);
+        assert_same_stringified(r#"b   	 +"c""#, r#""b +\"c\"""#);
+        assert_same_stringified(r#""+\\+\n+\"+""#, r#""\"+\\\\+\\n+\\\"+\"""#);
+        assert_same_stringified(r#""\'""#, r#""\"\\'\""#);  // TODO figure out why ' is different
+        // TODO test unicode chars
+    }
 }
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -1694,7 +1694,7 @@ int main(){}
         assert_unchanged("int \t\n\r     main() {}");
         assert_same_exact("int/* */main() {}", "int main() {}");
         assert_same_exact("int/*\n\n\n*/main() {}", "int\n\n\nmain() {}");
-        assert_same_exact("#define a(c) c\tc\na(1);a(2)", "\n1\t1;2\t2");
+        assert_same_exact("#define a(c) c\tc\na(1);a(2)", "\n1 1;2 2");
         assert_same_exact("#define a //\n#if defined a\n  x\n#endif", "\n\n  x\n");
         assert_same_exact("#define x\n#undef x\n  x", "\n\n  x");
         assert_same_exact("#pragma once\n  x", "\n  x");
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -1787,7 +1787,6 @@ h",
         assert_same_exact("#define hash #a\nhash", "\n#a");
     }
     #[test]
-    #[ignore = "Depends on #384"]
     fn stringify_string() {
         assert_same_stringified(r#""\'""#, r#""\"\\'\"""#);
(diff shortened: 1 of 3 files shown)


hdamron17:
Similarly to #395, this should maintain the original characters. I.e. `?` and `\x3f` should be handled differently even though they are identical semantically.

src/lex/replace.rs
```diff
@@ -312,3 +325,29 @@ fn replace_function(
...
+fn stringify(args: Vec<Token>) -> Token {
+    let ret: String = args
+        .into_iter()
+        .map(|arg| {
+            match arg {
+                Token::Whitespace(_) => String::from(" "), // Single space in replacement
+                Token::Literal(Literal::Str(s)) => {
+                    // TODO can we unwrap safely?
+                    let new_s = std::str::from_utf8(&s)
+                        .unwrap()
```
jyn514, line 338: Hmm, I thought this would fail on `"\xff"` but it seems to work reasonably well:
```c
$ cargo run -- -E
#define str(a) # a
str("\xff") 

 "�" 
```
I'm not sure why though.

src/lex/replace.rs
```diff
@@ -312,3 +325,29 @@ fn replace_function(
...
+                        .chars()
+                        .chain(new_s)
+                        .chain("\\\"".chars())
+                        .collect::<String>()
+                }
+                other => other.to_string(),
+            }
+        })
+        .collect();
+    Token::Literal(Literal::Str(ret.into()))
```
jyn514, line 352: This code looks right, but the behavior is wrong:
```c
#define str(a) # a
str(b)
// output
 b 
```
For comparison, clang outputs `"b"`.

src/lex/replace.rs
```diff
@@ -292,18 +292,31 @@ fn replace_function(
...
+                if let Some(index) = params.iter().position(|&param| param == id) {
+                    let replacement = args[index].clone();
+                    if next_stringify {
+                        replacements.push(stringify(replacement));
+                    } else {
+                        replacements.extend(replacement);
+                    }
+                } else {
+                    let token = Token::Id(id);
+                    replacements.push(token);
```
jyn514, line 309:
```suggestion
                    // only parameters can be stringified
                    let token = Token::Id(id);
                    replacements.push(token);
```

src/lex/replace.rs
```diff
@@ -292,18 +292,31 @@ fn replace_function(
...
+                }
+                next_stringify = true;
```
jyn514, line 311: Why do you unconditionally set this here? It seems like it would be wrong for e.g. `#define str(a) b a `.

src/lex/replace.rs
```diff
@@ -292,18 +292,31 @@ fn replace_function(
...
+            }
+            Token::Hash => {
+                next_stringify = true;
+            }
```
jyn514, line 315: Eventually it would be nice to support `##` as well. That's definitely a separate PR though.

src/lex/replace.rs
```diff
@@ -292,18 +292,31 @@ fn replace_function(
...
+        match *token {
+            Token::Id(id) => {
+                // #define f(a) { a + 1 } \n f(b) => b + 1
+                if let Some(index) = params.iter().position(|&param| param == id) {
+                    let replacement = args[index].clone();
+                    if next_stringify {
+                        replacements.push(stringify(replacement));
+                    } else {
+                        replacements.extend(replacement);
+                    }
```
jyn514, line 306:
```suggestion
                    // #define str(a) # a
                    if next_stringify {
                        replacements.push(stringify(replacement));
                    } else {
                        replacements.extend(replacement);
                    }
```

src/lex/replace.rs
jyn514: Ah, I found the issue - this should ignore whitespace tokens. For comparison, this works fine:
```c
#define f(a) #a
f(a)   
// output
"a"
```

src/lex/replace.rs
jyn514: Without the whitespace, it crashes as expected:
```c
#define str(a) #a 
str("\xff") 
Message:  called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 0, error_len: Some(1) }
Location: src/lex/replace.rs:337
```

jyn514:
> Similarly to #395, this should maintain the original characters. I.e. ? and \x3f should be handled differently even though they are identical semantically.

This will be tricky since the lexer currently parses hex escapes immediately. We may have to wait until https://github.com/jyn514/rcc/issues/384 is implemented.

jyn514:
Also, please add all the examples I gave as unit tests.

src/lex/replace.rs
hdamron17: I was indeed wrong. Theoretically it didn't matter because there can never be two Ids immediately together. I changed this when adding `pending_hash`.

src/lex/replace.rs
hdamron17: Another day.

src/lex/replace.rs
hdamron17: How do you suggest I handle this? I'd like to avoid writing my own string escape if possible.

hdamron17:
Everything except the issue of `str::from_utf8` failing is handled.

src/lex/cpp.rs
```diff
@@ -1644,4 +1644,31 @@ int main(){}
...
+        assert_same_stringified(r#"b   	 +"c""#, r#""b +\"c\"""#);
+        assert_same_stringified(r#""+\\+\n+\"+""#, r#""\"+\\\\+\\n+\\\"+\"""#);
+        assert_same_exact("#define xstr(a, b) #a = b\nxstr(1+2,3)", "\n\"1+2\" = 3");
+        assert_same_exact("#define xstr(a, b) a b\nxstr(1+2,3)", "\n1+2 3");
+        assert_same_exact("#define xstr(a) # a\nxstr(1+2)", "\n\"1+2\"");
+        assert_same_exact("#define hash #a\nhash", "\n#a");
+    }
+    #[test]
+    #[ignore = "Depends on #384"]
+    fn stringify_string() {
```
hdamron17, line 1666: These tests can be brought in after #384 is implemented.

src/lex/replace.rs
jyn514: This would also be helped by https://github.com/jyn514/saltwater/issues/384 I think - instead of greedily turning escapes into the relevant binary, we could instead stringify before processing escape sequences. I think without that stringifying doesn't make much sense anyway since you don't have a clear way to represent binary as text.

src/lex/cpp.rs
jyn514: Can you change `ignore` to `should_panic`? That way we'll have to update the test when it's been fixed.

hdamron17: It doesn't panic but just gives the wrong output.

src/lex/replace.rs
hdamron17: So should I add this to the `stringify_string` test and we'll handle it later?

jyn514: Yeah, that sounds reasonable.

src/lex/cpp.rs
jyn514: Ok, this is fine for now then.

src/lex/cpp.rs
```diff
@@ -1738,4 +1738,31 @@ h",
...
+    fn assert_same_stringified(cpp: &str, cpp_src: &str) {
+        assert_same_exact(
+            &format!("#define xstr(a) #a\nxstr({})", cpp),
+            &format!("\n{}", cpp_src),
+        );
+    }
+    #[test]
+    fn stringify() {
+        assert_same_stringified("a + b", r#""a + b""#);
+        assert_same_stringified(r#"b   	 +"c""#, r#""b +\"c\"""#);
```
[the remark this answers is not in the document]
hdamron17: The outer “ are not included in the actual string so the tokens on the left are `b`, `+` and `”c”`. So the resulting string is “b + \”c\””.

src/lex/replace.rs
```diff
@@ -329,18 +329,37 @@ fn replace_function(
         }
     }
 
+    let mut pending_hash: Vec<Token> = Vec::new(); // Holds hash until stringify
```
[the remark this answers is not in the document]
hdamron17: I know I had a reason but I don’t remember what it was now 😂

hdamron17:
No worries- I have to finish the rc-token branch first anyway so I’ll be occupied for a while.

src/lex/replace.rs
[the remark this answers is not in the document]
hdamron17: I believe this is to make sure we don’t lose the white space after the hash if a replacement does not happen.

hdamron17:
Turns out only " and \ are escaped as per [6.10.3.2](http://port70.net/~nsz/c/c11/n1570.html#6.10.3.2). This and a few other issues are fixed now.

r? @jyn514

src/lex/replace.rs
```diff
@@ -349,3 +371,26 @@ fn replace_function(
         .chain(replacements.into_iter().map(|t| Ok(location.with(t))))
         .collect()
 }
+
+fn stringify(args: Vec<Token>) -> Token {
+    let escape = |s: &str| s.replace('\\', "\\\\").replace('"', "\\\"");
```
jyn514, line 376:
```suggestion
    let escape = |s: &str| s.replace('\\', r#"\\"#).replace('"', r#"\""#);
```

src/lex/replace.rs
```diff
@@ -329,18 +332,37 @@ fn replace_function(
...
+            Token::Hash => {
+                pending_hash.push(token.clone());
+            }
+            _ => {
+                if !pending_hash.is_empty() && matches!(token, Token::Whitespace(_)) {
+                    pending_hash.push(token.clone());
+                } else {
+                    replacements.append(&mut pending_hash); // Dump pending_hash into replacements
+                    replacements.push(token.clone());
+                }
```
jyn514, line 364: If `!pending_hash.is_empty()` this should instead be an error.

> Each # preprocessing token in the replacement list for a function-like macro shall be followed by a parameter as the next preprocessing token in the replacement list.

src/lex/replace.rs
```diff
@@ -329,18 +332,37 @@ fn replace_function(
...
+                    if pending_hash.is_empty() {
+                        replacements.extend(replacement);
+                    } else {
+                        // #define str(a) #a
+                        replacements.push(stringify(replacement));
+                    }
+                } else {
+                    // only parameters can be stringified
+                    let token = Token::Id(id);
+                    replacements.push(token);
```
jyn514, line 351: This should be an error.

> Each # preprocessing token in the replacement list for a function-like macro shall be followed by a parameter as the next preprocessing token in the replacement list.

hdamron17:
I made your changes and added a few more tests. If you have anything specific you want, let me know.

r? @jyn514

src/lex/replace.rs
```diff
@@ -273,6 +273,16 @@ fn replace_function(
     let mut current_arg = Vec::new();
     let mut nested_parens = 1;
 
+    fn strip_whitespace(mut args: Vec<Token>) -> Vec<Token> {
+        if matches!(args.last(), Some(Token::Whitespace(_))) {
+            args.pop();
+        }
```
jyn514, line 279: !!!
```suggestion
        }
        assert!(!matches!(args.last(), Some(Token::Whitespace(_)));
```

src/lex/replace.rs
```diff
@@ -273,6 +273,16 @@ fn replace_function(
...
+        if matches!(args.first(), Some(Token::Whitespace(_))) {
+            args.remove(0);
+        }
```
jyn514, line 282:
```suggestion
        }
        assert!(!matches!(args.first(), Some(Token::Whitespace(_)));
```

hdamron17: Done

r? @jyn514

PR #460, opened 2020-05-30
Rename rcc to saltwater

jyn514:
# Why rename rcc?

rcc is a great name for a Rust C compiler. Unfortunately,
that binary name is already being used by the QT project for their
resource compiler: https://doc.qt.io/qt-5/rcc.html. This causes a lot of
confusion when people run `rcc` with QT installed.

 # Why saltwater?

I didn't want to go with synonyms for 'clang' because this is a separate
project which reuses no code or libraries from LLVM. There is an
informal tradition in Rust for having somewhat clever names, so I chose
'saltwater' since it's part of the sea (C) and is corrosive (causes rust).

 # Saltwater is pretty long to type, is that really the new binary name?

No, the binary name is `swcc` (and eventually `swcpp`).

Comments and suggestions for alternative names are welcome.

cc @Byter09 @pythondude325 @repnop @hdamron17 @xtachyon @Kixiron @memoryruins

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,7 +42,7 @@ Writing unit tests for codegen is not really possible,
 since we want to see how the program behaves at runtime,
 and running arbitrary C code in-process is a recipe for disaster.
 
-Instead, rcc uses 'Runner tests', which are C files in
+Instead, saltwater uses 'Runner tests', which are C files in
 `tests/runner-tests` that are compiled and run by `tests/runner.rs`.
 You can control the expected output using a comment at the top of the file
 (usually something like `// succeeds` or `code: 1`).
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "rcc"
+name = "saltwater"
 version = "0.9.0"
 authors = [
   "Joshua Nelson <jyn514@gmail.com>",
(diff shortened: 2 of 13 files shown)


jyn514:
Other names I thought of but were taken:
- [`crust`](https://docs.rs/crust/)

Other names I thought of/were suggested to me but rejected:
- `clatter`/`gong` (synonym to clang) - I don't want to tie this project to clang in people's minds
- `deadc` (because it's really salty) - This sounds like we're trying to kill the C language
- `recc`/`racc`/`ryck` (just add a vowel) - not obvious what the acronym stands for/how to pronounce it
- `cancer` (like the horoscope) - just no
- `ncc` (Nelson C Compiler) - I don't want to have my name on the project, that feels a little self-aggrandizing
- `Brachyura` (type of crab species) - nobody knows what this means, also it's not related to a C compiler
- `cling` (Rust is clinging to the past) - this is clever but I don't want to disparage C that much
- `rustc` :thinking: 
- `cc-rs` - not imaginative enough

hdamron17:
It’s easy to change later so no hurry to do anything now, but I’m not a fan of the phrasing “saltwater: the part of the sea causing lots of rust”. Also, is it possible to have the namespace `sw` instead of `saltwater` or is that too short and likely to cause conflicts?

jyn514:
> I’m not a fan of the phrasing “saltwater: the part of the sea causing lots of rust”

What would you suggest instead? I thought it was funny but I might be overestimating my own cleverness.

>  Also, is it possible to have the namespace sw instead of saltwater or is that too short and likely to cause conflicts?

`sw` is [already taken on crates.io](https://docs.rs/sw). I also don't think `saltwater` is too long a namespace for a library, only for binaries.

Kixiron:
For what it's worth, I think saltwater is a great name, it's memorable and unique. It's miles better than any other compiler I've seen

pythondude325:
We also need to edit the names of the discord and the messages in the `#welcome` channel.

jyn514:
Here is a full list of everything that will need to change:
- [ ] Code, including README, etc. (done in this PR)
- [ ] crates.io
    - [ ] Publish `saltwater 0.10.0`
    - [ ] Publish a new blank release of rcc pointing to saltwater
- [ ] discord
    - [ ] server name
    - [ ] welcome messages, links, etc.
- [ ] GitHub repo

jyn514:
Welcome to a new and glorious era of bad puns!

jyn514:
The rename is officially complete: there is a new version of saltwater published (https://crates.io/crates/saltwater) and the old version had a dummy release redirecting to it: https://crates.io/crates/rcc

PR #461, opened 2020-06-01
Escape special characters in display

### LINKED ISSUE #451: "\n" is pretty printed with a literal newline
### Expected behavior
`char *s = "Hello, world\n";` should not be changed after pretty-printing.

### Code

```c
char *s = "Hello, world
";
```

Stupremee:
This PR will escape the special characters `\n`, `\r`, and `\t` in the `Display` implementation for a `Literal` node.

Resolves #451

### DIFF
--- a/src/data/lex.rs
+++ b/src/data/lex.rs
@@ -372,7 +372,7 @@ impl std::fmt::Display for Literal {
             UnsignedInt(u) => write!(f, "{}", u),
             Float(n) => write!(f, "{}", n),
             Str(s) => {
-                let escaped = s
+                let mut escaped = s
                     .into_iter()
                     .flat_map(|c| match c {
                         b'\n' => "\\n".bytes().collect(),
@@ -381,6 +381,11 @@ impl std::fmt::Display for Literal {
                         _ => vec![*c],
                     })
                     .collect::<Vec<_>>();
+
+                // Remove the null byte at the end,
+                // because this will break tests and
+                // it's not needed in debug output.
+                escaped.pop();
 
                 write!(f, "\"{}\"", String::from_utf8_lossy(&escaped))
             }
@@ -456,5 +461,13 @@ pub(crate) mod test {
             let first = lexer.next().unwrap().unwrap().data;
             assert_eq!(&first.to_string(), *token);
         }
+    }
+
+    #[test]
+    fn str_display_escape() {
+        let token = "\"Hello, world\\n\"";
+        let mut lexer = cpp(token);
+        let first = lexer.next().unwrap().unwrap().data;
+        assert_eq!(&first.to_string(), token);
     }
 }
--- a/src/data/lex.rs
+++ b/src/data/lex.rs
@@ -465,7 +465,7 @@ pub(crate) mod test {
 
     #[test]
     fn str_display_escape() {
-        let token = "\"Hello, world\\n\"";
+        let token = "\"Hello, world\\n\\r\\t\"";
         let mut lexer = cpp(token);
         let first = lexer.next().unwrap().unwrap().data;
         assert_eq!(&first.to_string(), token);
--- a/src/data/lex.rs
+++ b/src/data/lex.rs
@@ -385,7 +385,7 @@ impl std::fmt::Display for Literal {
                 // Remove the null byte at the end,
                 // because this will break tests and
                 // it's not needed in debug output.
-                escaped.pop();
+                assert_eq!(escaped.pop(), Some(b'\0'));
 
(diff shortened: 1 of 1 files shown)


src/data/lex.rs
```diff
@@ -371,7 +371,19 @@ impl std::fmt::Display for Literal {
...
             UnsignedInt(u) => write!(f, "{}", u),
             Float(n) => write!(f, "{}", n),
-            Str(s) => write!(f, "\"{}\"", String::from_utf8_lossy(s)),
+            Str(s) => {
+                let escaped = s
+                    .into_iter()
+                    .flat_map(|c| match c {
+                        b'\n' => "\\n".bytes().collect(),
+                        b'\r' => "\\r".bytes().collect(),
+                        b'\t' => "\\t".bytes().collect(),
```
jyn514, line 380: I think this would be better as
```suggestion
                        b'\n' => b"\\n",
                        b'\r' => b"\\r",
                        b'\t' => b"\\t",
```

hdamron17:
Wouldn’t this depend on #384 since the newline should be printed in the same way that it was given (e.g. \n should be displayed \n but an actual newline in the string should be displayed as such)?

src/data/lex.rs
Stupremee: The problem is that `b""` returns a `&[u8; 2]` but `vec![*c]` has not the same type. We can also not use `&[*c]` or something similar because that would either has the wrong type, or would create a temporary value that is returned by the closure.

Stupremee:
My code only changes the `Display` implementation that is used in the `--debug-ast` option. I think it's much nicer to look at `\n` than an actual newline if you are inspecting the generated ast.

hdamron17:
Okay, as long as it’s never used outside of debugging, then I agree.

jyn514:
@hdamron17 newlines in string literals are illegal in C. Also, I dount stringification will be able to use the Display impl without changes anyway, I'd rather not handicap display for that purpose.

src/data/lex.rs
```diff
@@ -371,7 +371,24 @@ impl std::fmt::Display for Literal {
...
+                        _ => vec![*c],
+                    })
+                    .collect::<Vec<_>>();
+
+                // Remove the null byte at the end,
+                // because this will break tests and
+                // it's not needed in debug output.
+                escaped.pop();
```
jyn514, line 388:
```suggestion
                assert_eq!(escaped.pop(), b'\0');
```

src/data/lex.rs
jyn514: Hmm, that makes sense, and `[*c]` wouldn't work because it's length 1 instead of 2. I'd be interested if `[*c].as_slice()` would work but I bet that runs into lifetime issues.

src/data/lex.rs
```diff
@@ -445,4 +462,12 @@ pub(crate) mod test {
             assert_eq!(&first.to_string(), *token);
         }
     }
+
+    #[test]
+    fn str_display_escape() {
+        let token = "\"Hello, world\\n\\r\\t\"";
```
jyn514, line 468:
```suggestion
        let token = r#""Hello, world\n\r\t""#;
```

src/data/lex.rs
```diff
@@ -445,4 +462,12 @@ pub(crate) mod test {
...
+        let mut lexer = cpp(token);
+        let first = lexer.next().unwrap().unwrap().data;
+        assert_eq!(&first.to_string(), token);
```
jyn514, line 471: I forget exactly where to import this from (maybe `crate::parse::test`?), but this is the same thing more concisely. It also makes sure that only one token was lexed, not multiple.
```suggestion
        assert_no_change(token);
```

Stupremee: There are 3 `assert_no_change` methods, but they will all try to parse the input as source code. So I guess I will stick to `assert_eq`.

hdamron17:
Newlines may be illegal in c strings but tabs are legal and may arguably look nicer without being escaped. Although I see why all should be the same for consistency.

jyn514:
> tabs are legal

Oops, good point.

> may arguably look nicer without being escaped

I disagree, I think printing the tabs makes it ambiguous as to whether it's a tab or eight spaces. `\t` is unambiguous.

src/data/lex.rs
jyn514: I'll probably a version of this for literals in a later PR, but I agree for now `assert_eq` is fine.

jyn514:
Thanks for working on this!

PR #462, opened 2020-06-01
Skip whitespaces in preprocessor

### LINKED ISSUE #457: Space between function macro name and args
Using `-E`

### Expected behavior
Function macro should be replaced when whitespace separates name and args. (this is true in `clang` at least)
```c

<a>
```

### Actual Behavior
Name is consumed but the macro is not replaced.
```c

 (a)
```

### Code
```c
#define f(a) <a>
f (a)
```

Stupremee:
This PR will skip any white spaces that occur in the `lex::replace::replace` method, to support macro calls where
the identifier and the arguments are separated by spaces (e.g. `f (`).

Resolves #457

### DIFF
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -1647,7 +1647,9 @@ int main(){}
 
     #[test]
     fn space_separated_function_macro() {
-        let src = "#define f(a) <a>\nf     (a)";
-        assert_same(src, "<a>");
+        assert_same_exact("#define f(a) <a>\nf     (a)", "\n<a>");
+        assert_same_exact("#define f(a) <a>\nf(a)", "\n<a>");
+        assert_same_exact("#define f <a>\nf", "\n<a>");
+        assert_same_exact("#define f <a>\nf   ", "\n<a>   ");
     }
 }
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -1653,6 +1653,7 @@ int main(){}
         assert_same_exact("#define f(a) <a>\nf   ", "\nf   ");
         assert_same_exact("#define f(a) <a>\nf   ;", "\nf   ;");
         assert_same_exact("#define f(a) <a>\nf;", "\nf;");
+        assert_same_exact("#define f(a) 1\n#define h f (2)\nh", "\n1");
     }
 
     #[test]
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -1644,4 +1644,10 @@ int main(){}
         );
         // TODO test for #includes
     }
+
+    #[test]
+    fn space_separated_function_macro() {
+        let src = "#define f(a) <a>\nf     (a)";
+        assert_same(src, "<a>");
+    }
 }
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -1649,7 +1649,20 @@ int main(){}
     fn space_separated_function_macro() {
         assert_same_exact("#define f(a) <a>\nf     (a)", "\n<a>");
         assert_same_exact("#define f(a) <a>\nf(a)", "\n<a>");
-        assert_same_exact("#define f <a>\nf", "\n<a>");
-        assert_same_exact("#define f <a>\nf   ", "\n<a>   ");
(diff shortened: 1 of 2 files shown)


jyn514:
r? @hdamron17 

I'm worried this will lose whitespace again while printing with `-E` but I'll defer to Hunter's judgement.

jyn514:
As an alternative you could store the whitespace tokens and yield those tokens if no replacement was found.

Stupremee:
That sounds like a good idea

hdamron17:
As long as the macro is actually replaced then it is fine to consume the whitespace. The issue is if you have an id that is the name of a function macro but has is not used as one because it is not followed by a parenthesis. Can you `peek` at the token following the whitespace?

hdamron17:
Make a test using `assert_same_exact` that has both cases of `f(...)` being replaced and `f` left as is.

src/lex/cpp.rs
```diff
@@ -1647,7 +1647,9 @@ int main(){}
 
     #[test]
     fn space_separated_function_macro() {
-        let src = "#define f(a) <a>\nf     (a)";
-        assert_same(src, "<a>");
+        assert_same_exact("#define f(a) <a>\nf     (a)", "\n<a>");
+        assert_same_exact("#define f(a) <a>\nf(a)", "\n<a>");
+        assert_same_exact("#define f <a>\nf", "\n<a>");
+        assert_same_exact("#define f <a>\nf   ", "\n<a>   ");
```
jyn514, line 1653: This doesn't test the right thing.
```suggestion
        assert_same_exact("#define f(a) <a>\nf", "\nf");
        assert_same_exact("#define f(a) <a>\nf   ", "\nf   ");
```

hdamron17: Maybe add one more with something other than a parenthesis or eof after. E.g.
```
#define f(a) <a>
f   ;
```

src/lex/replace.rs
```diff
@@ -232,8 +223,42 @@ fn replace_function(
...
+                }
+                let id_token = Ok(location.with(Token::Id(id)));
+                errors.push(id_token);
+                errors.push(spaces);
+                return errors;
+            }
+            // If this branch is matched, this is not a macro call,
+            // since all other cases are covered above.
+            // So just append the identifier and the current token to the stack.
+            Some(Ok(Locatable { data: _, .. })) => {
```
jyn514, line 249:
```suggestion
            Some(Ok(_)) => {
```

src/lex/replace.rs
```diff
@@ -232,8 +223,42 @@ fn replace_function(
...
+                data: Token::Whitespace(_),
+                ..
+            })) => {
+                let spaces = inner.next().unwrap();
+                if let Some(Ok(Locatable {
+                    data: Token::LeftParen,
+                    ..
+                })) = inner.peek()
+                {
+                    inner.next();
```
jyn514, line 238: All of these calls to `inner` are incorrect. They should first look at `incoming` and only then fall back to `inner`. The different is that `incoming` takes into account tokens from macro replacements. As a result, your change handles this incorrectly:
```c
#define f(a) 1
#define h f (2)
h
```
The expected replacement is `1`, you output `f\n (2)`.

jyn514:
Can you add the example I gave as a unit test (or runner test if that's easier)?
```
// should be 1
#define f(a) 1
#define h f (2)
h
```

Stupremee:
The test is still failing with the following error:
```
left: `"\n\n1\n"`,
right: `"\n1\n"`',
```
I can't figure out why so you have to look over it.

src/lex/replace.rs
```diff
@@ -246,8 +249,8 @@ fn replace_function(
             // If this branch is matched, this is not a macro call,
             // since all other cases are covered above.
             // So just append the identifier and the current token to the stack.
-            Some(Ok(Locatable { data: _, .. })) => {
-                let token = inner.next().unwrap();
+            Some(Ok(_)) => {
+                let token = incoming.pop_front().or_else(|| inner.next()).unwrap();
```
hdamron17, line 253: While this is changing, I'd suggest making a lambda for `incoming.pop_front().or_else(|| inner.next())` since it's used thrice now and it'll reduce confusion going forward.

Issue #389, opened 2020-05-01
Add C highlighter to VSCode using semantic highlighting


jyn514:
This is what rust-analyzer does for Rust. Great way to test that my 'location' spans are correct.

https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide

cc @kixiron

PR #464, opened 2020-06-02
[WIP] Add a proper REPL

### LINKED ISSUE #372: Use rustyline for REPL

Stupremee:
This PR will introduce a proper Repl using [rustyline](https://crates.io/crates/rustyline) and the `jit` backend.
My current plan is to have a buffer which holds the code.
Every line will be processed like this:
- append code to code `buffer`
- check semantics of `buffer`
- if new code contains `main` method, execute it and reset the `buffer`
- otherwise just add the code to the `buffer`

It's probably possible to not reset the `buffer` and just override the `main` method in the existing program.
I'n also thinking about making the `buffer` not a string, but a list of parsed statements.

Resolves #372


src/lib.rs
```diff
@@ -87,6 +87,7 @@ pub mod intern;
 mod ir;
 mod lex;
 mod parse;
+pub mod repl;
```
jyn514, line 90: This should be part of the binary, not the library.

Cargo.toml
```diff
@@ -38,6 +38,8 @@ color-backtrace = { version = "0.4", default-features = false, optional = true }
 counter = "0.4"
 atty = { version = "0.2", default-features = false, optional = true } 
 git-testament = { version = "0.1", optional = true }
+rustyline = "6.1.2"
+rustyline-derive = "0.3.1"
```
jyn514, line 42: I would prefer for this to be an optional feature activated by default. That allows projects wanting to use rcc programmatically not to have to compile `rustyline`.
```suggestion
rustyline = { version = "6.1.2", optional = true }
rustyline-derive = { version = "0.3.1", optional = true }
```

You'll also have to add a `repl` feature under `[features]`, then add that features to `default = [ ... ]`.

src/repl/helper.rs
```diff
@@ -0,0 +1,71 @@
...
+    type Candidate = Pair;
+
+    fn complete(
+        &self,
+        _line: &str,
+        _pos: usize,
+        _ctx: &Context<'_>,
+    ) -> Result<(usize, Vec<Pair>), ReadlineError> {
+        Ok((0, vec![]))
+    }
```
jyn514, line 29: I'd love to do lots of fancy things with tab-completion, but that's a larger project, no need for it in a first draft.

src/repl/helper.rs
```diff
@@ -0,0 +1,71 @@
...
+    validate::{self, MatchingBracketValidator, Validator},
+    Context,
+};
+use rustyline_derive::Helper;
+use std::borrow::Cow;
+
+#[derive(Helper)]
+pub(super) struct ReplHelper {
+    pub(super) highlighter: MatchingBracketHighlighter,
+    pub(super) validator: MatchingBracketValidator,
```
jyn514, line 15: Hmm, this seems ok for the very first draft, but eventually I'd like to implement proper syntax highlighting.

src/repl/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+
+    pub fn run(&mut self) -> rustyline::Result<()> {
+        loop {
+            let line = self.editor.readline(self.prompt);
+            match line {
+                Ok(line) => self.process_line(line),
+                // Ctrl + C will do nothing
+                Err(ReadlineError::Interrupted) => continue,
```
jyn514, line 52: Will this stop the current line from being processed? For example
```
(bash@joshua-ThinkPad-P51s) ~/rcc ▶️ node
Welcome to Node.js v12.16.1.
Type ".help" for more information.
> function f() {
... 1 +  /* ctrl+c */
> 
```

src/repl/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+                // Ctrl + D will exit the repl
+                Err(ReadlineError::Eof) => std::process::exit(0),
```
jyn514, line 54: Better to `return Ok(())` which allows cleaning up resources along the way.

jyn514:
> if new code contains main method, execute it and reset the buffer, otherwise just add the code to the buffer

I'd actually like to evaluate things as expressions by default, not as declarations. This is how node behaves:
```
> 1 + 2
3
```
and also GHCI and python
```
Prelude> 1 + 2
3
>>> 1 + 2
3
```

You can parse things as expressions by creating a new Analyzer and calling `.expr()`. This would be easiest by using the `analyze` function in `src/analyze/mod.rs` - it's currently test-only, but I think it's fine to make it usable in all contexts.

src/repl/helper.rs
Stupremee: Since there's also #389, and probably other use cases for syntax highlighting, a generic module for syntax highlighting would be a good idea.

Stupremee:
But how should the result of the expression then be calculated? We do not have an interpreter, only a jit and I have no idea how we can get the result of the jitted code that gets executed.

src/bin/common/mod.rs
```diff
@@ -66,212 +65,12 @@ ARGS:
               Only one file at a time is currently accepted. [default: -]"
 );
 
-const USAGE: &str = "\
+pub(super) const USAGE: &str = "\
 usage: swcc [--help | -h] [--version | -V] [--debug-ir] [--debug-ast] [--debug-lex]
            [--debug-hir] [--jit] [--no-link | -c] [--preprocess-only | -E]
            [-I <dir>] [-D <id[=val]>] [<file>]";
```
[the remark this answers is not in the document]
Stupremee: Yes I realized that after I committed it :sweat_smile: 
It will be fixed asap.

src/analyze/mod.rs
```diff
@@ -1052,6 +1052,7 @@ impl PureAnalyzer {
     /// This returns an opaque index to the `Metadata`.
     fn declare(&mut self, mut decl: Variable, init: bool, location: Location) -> Symbol {
         if decl.id == "main".into() {
+            println!("main: {:?}", decl);
```
jyn514, line 1055:
```suggestion
```

src/bin/repl/mod.rs
```diff
@@ -97,6 +97,56 @@ impl<'s> Repl<'s> {
...
+        };
+
+        let span = expr.location;
+        let stmt = span.with(hir::StmtType::Return(Some(expr)));
+        let init = hir::Initializer::FunctionBody(vec![stmt]);
+        let decl = hir::Declaration {
+            symbol: main.insert(),
+            init: Some(init),
+        };
+        let (result, _warns) = ir::compile(module, vec![span.with(decl)], false);
```
jyn514, line 139: This looks useful on its own, can you make it a function in `src/lib.rs` called `eval()` that's implemented on `JIT`?

src/bin/repl/mod.rs
```diff
@@ -97,6 +97,56 @@ impl<'s> Repl<'s> {
...
+                inline: false,
+                no_return: false,
+            },
+        };
+
+        let main = hir::Variable {
+            ctype: types::Type::Function(function_type),
+            storage_class: data::StorageClass::Extern,
+            qualifiers,
+            id: saltwater::InternedStr::get_or_intern("main"),
```
jyn514, line 129:
```suggestion
            id: "main".into(),
```

src/bin/repl/mod.rs
```diff
@@ -97,6 +97,56 @@ impl<'s> Repl<'s> {
...
+            varargs: false,
+        };
+        let qualifiers = hir::Qualifiers {
+            volatile: false,
+            c_const: false,
+            func: hir::FunctionQualifiers {
+                inline: false,
+                no_return: false,
+            },
+        };
```
jyn514, line 123:
```suggestion
        let qualifiers = hir::Qualifiers::default();
```

src/bin/repl/mod.rs
```diff
@@ -97,6 +97,56 @@ impl<'s> Repl<'s> {
     }
 
     fn execute_code(&mut self, code: &str) {
-        let program = check_semantics(code, self.opt.clone());
+        let module = initialize_jit_module();
+        let expr = match analyze(code, Parser::expr, PureAnalyzer::expr) {
+            Ok(mut expr) => {
+                expr.ctype = types::Type::Int(true);
```
jyn514, line 103: This doesn't look right. What's this meant to do?

src/bin/repl/mod.rs
```diff
@@ -97,6 +97,56 @@ impl<'s> Repl<'s> {
...
+            init: Some(init),
+        };
+        let (result, _warns) = ir::compile(module, vec![span.with(decl)], false);
+
+        if let Err(err) = result {
+            println!("failed to compile: {}", err.data);
+            return;
+        }
+
+        let module = result.unwrap();
```
jyn514, line 146: Try to avoid `unwrap` if possible.
```suggestion
        let module = match result {
            Err(err) => {
                println!("failed to compile: {}", err.data);
                return;
            }
            Ok(module) => module),
        };
```

src/bin/repl/mod.rs
jyn514: Yeah, this crashes on strings:
```
>> "a"
Message:  verification error: - inst1 (return v0): arg 0 (v0) has type i64, must match function signature of i32

note: while compiling function u0:0() -> i32 system_v {
    gv0 = symbol colocated u1:0

block0:
    v0 = global_value.i64 gv0
    return v0
}
```

jyn514: Instead you could use a function name other than `main` and set the return type to the type of the expression. Then you decide how to display the result depending on the type. That would also fix the current bug where `'a'` is displayed as `99`.

src/analyze/mod.rs
Stupremee: ups :sweat_smile:

src/bin/repl/mod.rs
Stupremee: That's a good idea

Stupremee: I had this code there, because otherwise I got an error from cranelift that it expected an i32 but got i64.

Stupremee:
I will close this pull request and reopen one because I have lost the overview and am not really satisfied with this one.

PR #466, opened 2020-06-04
Add salty scream to panic

pythondude325:
Added the R2-D2 scream


src/main.rs
```diff
@@ -550,6 +550,22 @@ fn install_panic_hook() {
...
+        {
+            const SCREAM: &[u8] = include_bytes!("data/R2D2-Scream.wav");
+
+            if let Ok(mut child) = std::process::Command::new("aplay")
+                .stdin(std::process::Stdio::piped())
+                .stdout(std::process::Stdio::null())
+                .spawn()
+            {
+                use std::io::Write;
+                let stdin = child.stdin.as_mut().expect("Unable to open stdin");
```
Evrey, line 564: Just bail out, don't crash. Poor compiler-chan had enough pain already.

src/main.rs
```diff
@@ -550,6 +550,22 @@ fn install_panic_hook() {
         let mut rng = rand::thread_rng();
         let msg = msgs[rng.gen_range(0, msgs.len())];
         println!("{}", msg);
+
+        #[cfg(target_os = "linux")]
+        {
+            const SCREAM: &[u8] = include_bytes!("data/R2D2-Scream.wav");
+
+            if let Ok(mut child) = std::process::Command::new("aplay")
```
jyn514, line 558: Why use `aplay` directly instead of `rusty_audio`?

Evrey: Me guesses sacrificing cross-platformness for a low dependency count and simple, tiny code.

Kixiron: I vote for `rusty_audio` so that all OS's may witness this beauty

pythondude325:
use this command to test:
```sh
cargo run --features=salty <(echo ሰ)
```

src/main.rs
```diff
@@ -528,6 +528,15 @@ mod backtrace {
     }
 }
 
+fn play_scream() -> Option<()> {
+    const SCREAM: &[u8] = include_bytes!("data/R2D2-Scream.ogg");
+    let device = rodio::default_output_device()?;
+    let source = rodio::Decoder::new(std::io::Cursor::new(SCREAM)).ok()?;
+    rodio::play_raw(&device, rodio::source::Source::convert_samples(source));
+    std::thread::sleep(std::time::Duration::from_millis(2835));
```
jyn514, line 536: Is there no way to tell the length of the audio from `rodio`?

src/main.rs
```diff
@@ -528,6 +528,15 @@ mod backtrace {
     }
 }
 
+fn play_scream() -> Option<()> {
```
jyn514, line 531: Since you're using `None` to say there was an error, I think this should instead be `Result<(), ()>`.

jyn514, line 531:
```suggestion
#[cfg(feature = "salty")]
fn play_scream() -> Option<()> {
```
This will fail to compile otherwise.

src/main.rs
jyn514: Also, I'm undecided whether we should actually wait for it to finish, I think it might be funny to let the audio play while you read the panic.

pythondude325: There is but it doesn't work

PR #467, opened 2020-06-04
Added more GLaDOS quotes to warning, error, and panic messages.

1011X:


src/main.rs
```diff
@@ -446,6 +448,9 @@ fn error<T: std::fmt::Display>(msg: T, location: Location, file_db: &Files, colo
             // glados quotes
             "I'd just like to point out that you were given every opportunity to succeed",
             "Okay. Look. We both did a lot of things that you're going to regret. But I think we can put our differences behind us. To prevent segfaults. You monster.",
+            "Congratulations. Not on the code.",
+            "You know, if you'd done that to some other codebase, they might devote their existence to exacting revenge.",
+            "Not bad. I forgot how good you are at this. You should pace yourself, though. We have A LOT of code to compile.",
```
jyn514, line 453: Hmm, this seems a little odd for an error. I like the quote but I'm not sure it fits.

src/main.rs
```diff
@@ -205,6 +205,8 @@ fn handle_warnings(warnings: VecDeque<CompileWarning>, file_db: &Files, color: C
             "To UB or not to UB? That is … apparently a question you never ask yourself.",
             "Do you _really_ want to know what this compiles to?",
             "Not the type to ever stop and think for a moment, eh? Here's your moment, use it well.",
+            "Federal regulations require me to warn you that this next warning... is looking pretty good.",
+            "Try to avoid this garbage you're writing.",
```
jyn514, line 209: This automatically gets `:` appended after, so avoid periods unless it's multiple sentences.
```suggestion
            "Federal regulations require me to warn you that this next warning... is looking pretty good",
            "Try to avoid this garbage you're writing",
```

src/main.rs
1011X: Hm, okay, I'll leave it out for now and see if i can make it more fitting later

PR #468, opened 2020-06-05
Use chars instead of u8 for the lexer; don't panic on unicode identifiers

jyn514:
Closes https://github.com/jyn514/saltwater/issues/265

r? @pythondude325


jyn514:
Previously, `self.location.offset` held the last byte of the last token to be returned. For example, in `abc def`, after returning `Token::Id(abc)`, the offset would be `2`. Now, the offset holds the byte of the first character of the next token: 3.

As a result, whereas before we knew we were at the end of the file when we saw `self.location.offset = self.chars.len() - 1`, now we are at the file whenever the lengths are equal. In other words, we no longer have a 'first time we saw EOF' state. As a result, in order to check for missing newlines, there has to be an explicit check for it whenever we return the last token. This took me a while to debug.

jyn514:
Some useful follow-up PRs:
- Separate out the inner iterator, it's gotten pretty chunky and it would be useful semantically to separate the methods. In rustc this is [`cursor.rs`](https://github.com/rust-lang/rust/blob/master/src/librustc_lexer/src/cursor.rs).
- Allow unicode identifiers (finally!) This requires a dependency on https://docs.rs/unicode-xid/.

src/lex/mod.rs
```diff
@@ -318,29 +321,29 @@ impl Lexer {
...
                 }
             }
         };
-        if self.match_next(b'.') {
+        if self.match_next('.') {
             return self.parse_float(radix, buf).map(float_literal);
         }
-        if let Some(b'e') | Some(b'E') | Some(b'p') | Some(b'P') = self.peek() {
+        if let Some('e') | Some('E') | Some('p') | Some('P') = self.peek() {
             buf.push_str(".0"); // hexf doesn't like floats without a decimal point
```
pythondude325, line 341: I should probably fix this.

src/lex/cpp.rs
```diff
@@ -1061,14 +1062,10 @@ impl<'a> PreProcessor<'a> {
     // now we want to do the dirty work of reading it into memory
     fn include_path(
         &mut self,
-        filename: Vec<u8>,
+        filename: String,
```
pythondude325, line 1065: `filename` should probably be an `std::path::PathBuf` so you get easier path semantics.

src/lex/mod.rs
jyn514: This was actually written when I was still using hexf, I'm not sure if it's true for `hexponent`.

PR #470, opened 2020-06-07
Fix macro redefinition semantics

pythondude325:
Macro redefinition was broken. Added a new function to define macros for internal use and created an error for the erroneous macro redefinition.

### DIFF
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -488,6 +488,11 @@ pub enum CppError {
     /// After parsing an `#if` expression, there were tokens left over.
     #[error("trailing tokens in `#if` expression")]
     TooManyTokens,
+
+    /// If a macro is redefined, its definition must be identical to as it's old
+    /// definition.
+    #[error("redefined macro definition for '{0}' does not match original definition")]
+    IncompatibleRedefinition(InternedStr),
 }
 
 /// Lex errors are non-exhaustive and may have new variants added at any time
--- a/src/data/error.rs
+++ b/src/data/error.rs
@@ -489,9 +489,9 @@ pub enum CppError {
     #[error("trailing tokens in `#if` expression")]
     TooManyTokens,
 
-    /// If a macro is redefined, its definition must be identical to as it's old
-    /// definition.
-    #[error("redefined macro definition for '{0}' does not match original definition")]
+    /// If a macro is redefined, the new definition must be identical to the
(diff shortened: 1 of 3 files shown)


src/data/error.rs
```diff
@@ -488,6 +488,11 @@ pub enum CppError {
     /// After parsing an `#if` expression, there were tokens left over.
     #[error("trailing tokens in `#if` expression")]
     TooManyTokens,
+
+    /// If a macro is redefined, its definition must be identical to as it's old
+    /// definition.
+    #[error("redefined macro definition for '{0}' does not match original definition")]
+    IncompatibleRedefinition(InternedStr),
```
jyn514, line 495: I think you a word.

```suggestion
    /// If a macro is redefined, the new definition must be identical to the original.
    #[error("redefinition of '{0}' does not match original definition")]
    IncompatibleRedefinition(InternedStr),
```

src/lex/cpp.rs
```diff
@@ -923,16 +923,34 @@ impl<'a> PreProcessor<'a> {
...
     }
+    fn define_macro(&mut self, name: InternedStr, definition: Definition) -> Result<(), CppError> {
+        use std::collections::hash_map::Entry;
+        match self.definitions.entry(name) {
+            Entry::Vacant(entry) => {
+                entry.insert(definition);
+                Ok(())
+            }
+            Entry::Occupied(entry) => {
+                // https://port70.net/~nsz/c/c11/n1570.html#6.10.3p1
```
jyn514, line 945: Let's just put the section and not the full URL.
```suggestion
                // 6.10.3p1
```

src/lex/cpp.rs
```diff
@@ -1417,13 +1435,28 @@ a";
...
+            CppError::IncompatibleRedefinition(_),
+            "incompatible redefinition"
+        );
+
+        let src = "
+#define a b
+#define a b
+a
+";
+        assert_same(src, "b");
```
jyn514, line 1459: Can you also add a couple tests for redefining a function macro?

pythondude325: Look 9 lines down ;)

src/lex/cpp.rs
pythondude325: Look 9 lines down ;)

PR #474, opened 2020-06-12
Close #384: Return only span and type from lexer

### LINKED ISSUE #384: The lexer should only return a span and a token-kind

hdamron17:


hdamron17:
@jyn514 Let me know what you think so far before I get too far into any rabbit holes.

hdamron17:
Right now the whole span is dumped at the end, but I am planning to separate out a `LiteralParser` trait which will contain all the `parse_*` functions from `Lexer` and will be implemented for both `Lexer` and `Peekable<char>`. Unfortunately, it will require constants to be parsed twice- once for determining their type in the lexer and again in codegen. But the second parse (in theory) should be simple and error-free because it's already been done once.

src/data/lex.rs
```diff
@@ -147,7 +159,7 @@ pub enum Literal {
     Int(i64),
     UnsignedInt(u64),
     Float(f64),
-    Str(Vec<u8>),
+    Str,
```
jyn514, line 162: Yay, we can finally `derive(Copy)` on Token!

src/ir/mod.rs
```diff
@@ -391,6 +397,9 @@ impl<B: Backend> Compiler<B> {
 
         Ok(())
     }
+    fn resolve_location(&self, loc: Location) -> &str {
+        self.files.source_slice(loc.file, loc.span).unwrap() // TODO unwrap safely
```
jyn514, line 401: I just had an idea: if we made the fields on Location private we could ensure all locations came from the lexer, and so are always in bounds (modulo bugs).

src/ir/static_init.rs
```diff
@@ -112,11 +112,9 @@ impl<B: Backend> Compiler<B> {
...
         })
     }
-    pub(super) fn compile_string(
-        &mut self,
-        string: Vec<u8>,
-        location: Location,
-    ) -> CompileResult<DataId> {
+    pub(super) fn compile_string(&mut self, location: Location) -> CompileResult<DataId> {
+        let string = self.resolve_location(location);
+        let string: Vec<u8> = string.get(1..string.len() - 1).unwrap().as_bytes().to_vec(); // Remove quotation marks // TODO escapes?
```
jyn514, line 117: RE escapes: this seems like something we might have to calculate in various different areas. Instead of redoing it every time, we could have a query infrastructure like rustc using [salsa](https://docs.rs/salsa/0.14.1/salsa/). That also gets us incremental compilation almost for free.

cc @pythondude325

hdamron17:
> > I am planning to separate out a LiteralParser trait which will contain all the parse_* functions from Lexer and will be implemented for both Lexer and Peekable.
> 
> I'm not sure I quite understand the use case? Why have a trait?

This would just be to avoid duplicating the parse code. So we can do parsing from the real lexer and from a pseudo-lexer in the codegen stage.

hdamron17:
I still need to change the size of allocated memory for strings, but other than that it seems to be working. Some tests are broken, though, because they rely on `Literal::Str` storing the string.

hdamron17:
I rebased onto your big changes in master and it fixes consecutive strings with newline here. There's a backup of the previous version at https://github.com/hdamron17/saltwater/tree/lexer-span-bak too just in case I botched the rebase. Can you check that everything looks kosher? @jyn514

src/data/lex.rs
```diff
@@ -15,6 +15,18 @@ pub struct Span {
     pub end: u32,
 }
 
+impl Span {
+    pub fn len(&self) -> usize {
+        (self.end - self.start) as usize // TODO is this unsafe?
```
jyn514, line 20: I assume by 'unsafe' you mean 'can panic'? Rust usually uses unsafe to mean memory corruption and that sort of thing.

src/ir/static_init.rs
```diff
@@ -422,7 +423,10 @@ impl Literal {
                     x, f
                 )),
             }),
-            Literal::Str(string) => Ok(string.into_boxed_slice()),
+            Literal::Str(_) => {
+                todo!("string literal")
```
jyn514, line 427: Maybe something like this?
```rust
Ok(self.literal_parser_iter(expr.location)
    .parse_string_raw()
    .unwrap()
    .into_boxed_slice())
```

src/lex/mod.rs
```diff
@@ -923,3 +600,365 @@ impl Iterator for Lexer {
...
+        } else {
+            false
+        }
+    }
+    /// Read a logical character, which may be a character escape.
+    ///
+    /// Has a side effect: will call `warn` if it sees an invalid escape.
+    ///
+    /// Before: u8s{"\b'"}
+    /// After:  chars{"'"}
```
jyn514, line 642: Whoops, this is left over from a bad change in https://github.com/jyn514/saltwater/pull/468 
```suggestion
    /// Before: chars{"\b'"}
    /// After:  chars{"'"}
```

src/ir/static_init.rs
hdamron17: Problem is this is in `impl Literal` not `Compiler` so I'd have to pass in `&Compiler`. Not terribly inconvenient.

src/data/lex.rs
hdamron17: Yeah, that's what I mean. Is this reasonable to do?

src/ir/static_init.rs
hdamron17: I moved `into_bytes` into `impl Compiler` and it works. Hopefully that's not an issue in the eyes of the API.

src/data/lex.rs
jyn514: Yeah, the implementation looks fine.

src/lex/mod.rs
```diff
@@ -646,6 +440,211 @@ pub(crate) trait LiteralParser {
...
+            if c.is_digit(radix.as_u8().into()) {
+                self.next_char();
+                buf.push(c);
+            } else {
+                break;
+            }
+        }
+        // in case of an empty mantissa, hexf doesn't like having the exponent right after the .
+        // if the mantissa isn't empty, .12 is the same as .120
+        //buf.push('0');
```
jyn514:
```suggestion
```

src/lex/mod.rs
```diff
@@ -646,6 +440,211 @@ pub(crate) trait LiteralParser {
...
+                } else {
+                    return Err(LexError::MissingDigits(radix.try_into().unwrap()));
+                }
+            }
+        };
+        if self.match_next('.') {
+            return self.parse_float(radix, buf).map(float_literal);
+        }
+        if let Some('e') | Some('E') | Some('p') | Some('P') = self.peek() {
+            buf.push_str(".0"); // hexf doesn't like floats without a decimal point
```
jyn514, line 497: I never cleaned up after the hexponent migration, so now you get to do it :P
```suggestion
```

src/data/lex.rs
```diff
@@ -152,14 +164,13 @@ pub enum ComparisonToken {
...
-#[derive(Copy, Clone, Debug, PartialEq)]
-#[cfg_attr(test, derive(Arbitrary))]
+#[derive(Clone, Debug, PartialEq)]
 pub enum Literal {
     // literals
     Int(i64),
     UnsignedInt(u64),
     Float(f64),
-    Str(usize), // Stores length of string
+    Str(Vec<RcStr>, usize), // second arg is length of parsed string
```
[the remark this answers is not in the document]
hdamron17: The RcStr is the source code and the length is after parsing (so we know how big of an array to declare).

hdamron17:
Much of it is changing because I have to move it to the LiteralParser trait. I didn’t see a good way to make the diff less obnoxious.

src/lex/tests.rs
```diff
@@ -46,7 +46,28 @@ where
 }
 
 fn match_char(lexed: Option<LexType>, expected: u8) -> bool {
-    match_data(lexed, |c| c == Ok(&LiteralToken::Char(expected).into()))
+    use crate::data::hir::LiteralValue;
+    match lexed {
```
hdamron17, line 50: @jyn514 pls help

src/data/lex.rs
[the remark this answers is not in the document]
jyn514: Oh I see, this is the length after we parse escapes.

src/lex/tests.rs
```diff
@@ -46,7 +46,28 @@ where
...
+                match loc.parse().data {
+                    LiteralValue::Char(c) => c == expected,
+                    _ => false,
+                }
+            } else {
+                return false;
+            }
+        }
+        _ => false,
+    }
```
jyn514, line 70: Implementing `.parse()` on the `Locatable` is a little odd, can you implement it on `LiteralToken` instead?

Also, why do you match on the value after parsing it? Shouldn't it always be a `Char`? It might be cleaner to implement a standalone parsing function that doesn't go through `LiteralValue`.

```suggestion
    match lexed {
        Some(Ok(Locatable {
            data: Token::Literal(literal @ LiteralToken::Char(_)),
            location,
        } =>  match literal.parse() {
                    LiteralValue::Char(c) => c == expected,
                    _ => false,
                }
        _ => false,
    }
```

src/lex/mod.rs
```diff
@@ -1039,7 +1019,31 @@ impl LiteralToken {
...
             LiteralToken::UnsignedInt(u) => LiteralValue::UnsignedInt(u),
-            LiteralToken::Float(f) => LiteralValue::Float(f),
+            LiteralToken::Float(rcstr) => {
+                let buf = rcstr.as_str();
+                let hex = buf.starts_with("0x");
+                let buf = buf
+                    .trim_end_matches("f")
+                    .trim_end_matches("F")
+                    .trim_end_matches("l")
+                    .trim_end_matches("L"); // Ignored for compatibility reasons
```
jyn514, line 1029: This does not have the same behavior. Previously `1.0fl` would be parsed as `[Float(1.0), Id("l")]`. Now it will be parsed as one float. This needs to stop the first time it sees one of the letters.

Unless this only works on input that's already valid? In that case it should be fine.

hdamron17: This is operating on a string after it has already been lexed so it shouldn't be able to have more than one suffix, but I'll try to do it correctly. Unfortunately `strip_suffix` is nightly which is the best way I can see to do it.

src/lex/mod.rs
```diff
@@ -375,281 +220,80 @@ impl Lexer {
...
+    fn parse_int(&mut self, radix: Radix) -> Result<Option<()>, LexError> {
         let parse_digit = |c: char| match c.to_digit(16) {
             None => Ok(None),
             Some(digit) if digit < radix.as_u8().into() => Ok(Some(digit)),
             // if we see 'e' or 'E', it's the end of the int, don't treat it as an error
-            // if we see 'b' this could be part of a binary constant (0b1)
-            // if we see 'f' it could be a float suffix
             // we only get this far if it's not a valid digit for the radix, i.e. radix != 16
-            Some(11) | Some(14) | Some(15) => Ok(None),
+            Some(14) => Ok(None),
```
hdamron17, line 255: Can you explain when the b and f could be there?

hdamron17:
I think this is almost ready to ship out. Some thoughts:
* There are some TODOs in the code, mostly thought provoking, where comments are welcome.
* Can I just merge master into this branch instead of rebasing, just this once?
* What additional tests should I provide?

hdamron17:
Oops I accidentally merged already trying to figure out GitHub's web editor.

src/lex/mod.rs
```diff
@@ -171,7 +171,7 @@ impl Lexer {
         if self.parse_int(radix)?.is_none()
             && !(radix == Radix::Octal || radix == Radix::Decimal || self.peek() == Some('.'))
         {
-            return Err(LexError::MissingDigits(radix.try_into().unwrap()));
+            return Err(LexError::MissingDigits(radix));
```
jyn514, line 174: This was fixed on master :/ if you rebase it will fix the rest of the clippy lints.

hdamron17:
shared_str merged my pull request, but they still need to publish a new version, right? Is it reasonable to ask for that?

jyn514:
Yes, they'll need to publish to crates.io. I think that's fine to ask for.

src/lex/mod.rs
hdamron17: I tried rebasing (I just squashed everything together first because I move things several times, sometimes pointlessly) and still had to manually fix this one. Would you prefer the merge commit as it is or a rebased squash commit?

src/data/lex.rs
```diff
@@ -140,17 +163,31 @@ pub enum ComparisonToken {
...
-    Str(Vec<u8>),
-    Char(u8),
+    Int(RcStr),
+    UnsignedInt(RcStr),
+    Float(RcStr),
+    Str(Vec<RcStr>),
+    Char(RcStr),
 }
 
+impl PartialEq for LiteralToken {
```
hdamron17, line 176: The options for comparison are:
* Exact pointer location matches
* Source strings match (this implementation)
* The literal is parsed and we see if this is valid
Which is preferred? I'm not entirely sure how it is used.

src/lex/cpp.rs
```diff
@@ -687,9 +692,9 @@ impl<'a> PreProcessor<'a> {
                 } if name == defined => {
                     let def = Self::defined(&mut lex_tokens, location)?;
                     let literal = if definitions.contains_key(&def) {
-                        Literal::Int(1)
+                        LiteralToken::Int(RcStr::from("1")) // TODO can we not do tokens here?
```
hdamron17, line 695: This works, but there's probably a better way.

Cargo.toml
```diff
@@ -109,3 +110,6 @@ overflow-checks = false
 all-features = true
 # opt-out of non-default targets
 targets = []
+
+[patch.crates-io]
+shared_str = { git = 'https://github.com/hdamron17/shared_str.git' }
```
jyn514: I think this can be removed now 0.1.1 is published :)

src/lex/mod.rs
jyn514: `0b1110` or `1.431f`. The `e` is for exponents: `1.1e2`

src/lex/mod.rs
jyn514: Oh I missed that you merged because the diff is so big. Yeah that's fine.

src/lex/mod.rs
hdamron17: The 'b' is consumed at the beginning of a binary constant so that's invalid now, and we don't use `parse_int` for consuming after the decimal place to 'f' would always be invalid here. I think I see what you're doing though - we want to consume all the alpha characters after the digit and report it as an error so there's not a duplicate 'invalid identifier' error.

Cargo.toml
hdamron17: I'm pretty sure I did this and you're looking at an older version...

src/lex/mod.rs
jyn514: Hmm ... your reasoning makes sense but I remember I needed them for something. 59157678e9a94d1998d2dad4a04f7316d507fa53 and a05e14d6820ee3a5ba22cb5e3338370c0ca38646 are the relevant commits.

Isn't this method also used for the tail parts of floats? I thought it went `fn parse_num { let int = parse_int(); if match_next('.') { let fraction = parse_int(); } }`.

hdamron17: For the 'f' suffix, this was added when `parse_int` was used for digits after the decimal point but not anymore - [see here](https://github.com/jyn514/saltwater/commit/524c93d#diff-72dac0bc15578b7a7f1e37d8d65305dbR206).

For the 'b' in binary constants, I'm not sure why it was ever necessary because the 'b' was always removed as far as I can tell.

src/data/lex.rs
hdamron17: Cool cool

src/data/hir.rs
```diff
@@ -153,6 +153,17 @@ pub enum ExprType {
...
 }
 
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(test, derive(Arbitrary))]
+pub enum LiteralValue {
+    // literals
+    Int(i64),
+    UnsignedInt(u64),
+    Float(f64),
+    Str(Vec<u8>), // second arg is length of parsed string
```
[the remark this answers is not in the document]
hdamron17: Oh I forgot to remove the comment when I took it out. It turned out to not be necessary since `LiteralToken`s don't go all the way to codegen.

src/data/lex.rs
```diff
@@ -363,32 +397,23 @@ impl std::fmt::Display for Token {
...
-                // it's not needed in debug output.
-                assert_eq!(escaped.pop(), Some(b'\0'));
-
-                write!(f, "\"{}\"", String::from_utf8_lossy(&escaped))
-            }
-            Char(c) => write!(f, "'{}'", char::from(*c).escape_default()),
+                    .map(RcStr::as_str)
+                    .collect::<Vec<_>>()
+                    .join(" ")
+            ),
```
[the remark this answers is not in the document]
hdamron17: Much cleaner, thanks!

src/data/lex.rs
```diff
@@ -444,6 +460,38 @@ impl From<ComparisonToken> for Token {
...
+    use super::LiteralToken;
+    use proptest::prelude::*;
+    use shared_str::RcStr;
+
+    impl Arbitrary for LiteralToken {
+        type Parameters = ();
+        type Strategy = BoxedStrategy<LiteralToken>;
+        fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
+            prop_oneof![
+                // TODO give regex of all possible literals
```
jyn514, line 474: That would be a very painful regex, because we'd have to support decimal/hexadecimal floats as well.

src/lex/mod.rs
```diff
@@ -920,3 +567,498 @@ impl Iterator for Lexer {
...
+    }
+}
+
+pub(crate) trait LiteralParser {
+    fn next_char(&mut self) -> Option<char>;
+    fn peek(&mut self) -> Option<char>;
+    fn peek_next(&mut self) -> Option<char>;
+    fn get_location(&self) -> &SingleLocation;
+    fn handle_error(&mut self, err: Locatable<LexError>);
+    fn handle_warning(&mut self, err: Locatable<Warning>);
```
jyn514, line 591:
```suggestion
    fn err(&mut self, err: Locatable<LexError>);
    fn warn(&mut self, err: Locatable<Warning>);
```

src/lex/mod.rs
```diff
@@ -920,3 +567,498 @@ impl Iterator for Lexer {
...
+
+    fn handle_error_loc<E: Into<LexError>>(&mut self, err: E, location: Location) {
```
jyn514, line 593:
```suggestion
    fn err_loc<E: Into<LexError>>(&mut self, err: E, location: Location) {
```

src/lex/mod.rs
```diff
@@ -920,3 +567,498 @@ impl Iterator for Lexer {
...
+        self.handle_error(location.with(err.into()));
+    }
+    fn handle_warning_loc<W: Into<Warning>>(&mut self, warning: W, location: Location) {
```
jyn514, line 596:
```suggestion
    fn warn_loc<W: Into<Warning>>(&mut self, warning: W, location: Location) {
```

src/data/lex.rs
hdamron17: I'll leave the TODO in case someone feels froggy. I think it's not necessary unless we start using proptest to ensure we parse a token iff it's valid (how I feel like it should be used).

Issue #475, opened 2020-06-12
Get rid of unput in lexer


jyn514:
Currently, unput exists _only_ for this hack:

https://github.com/jyn514/saltwater/blob/546ed7de472c2be7b57c3e44fba628afb856b9ae/src/lex/mod.rs#L662

If we got rid of that, we could get rid of `unput()` altogether, which I've been trying to do for ages. This will make #474 significantly easier.

## Why does this exist?

The use case for `seen_line_token` is that `#warning x` is a preprocessor directive, but `1 + #warning x is not`. This is part of the reason I tied the preprocessor to the lexer. Say you have
```c
"a"
"b"
```
Those two tokens should be concatenated into a single string, in which case everything is fine. However, if you have
```c
"a"
# warning
```
then the preprocessor needs to know no token has been seen yet on the second line.
But `seen_line_token` gets set after `parse_string` returns!  So it's treated as if you wrote `"a" #warning`, which is not correct. The current hack is to put back a newline character if we saw a newline in `consume_whitespace`, which requires `unput`

## What is the fix?

We can't use `consume_whitespace_no_newline` because we need to know about following strings, there's no way to conditionally consume newlines.

Instead, we can change the algorithm: concatenate the strings in the preprocessor (or parser) instead of the lexer. The main difference there is that the preprocessor can store a pending token, where as the lexer can only store a pending character.

This should also solve https://github.com/jyn514/saltwater/issues/361.

jyn514:
Fixed in #478

PR #477, opened 2020-06-13
Get rid of `first` argument to Parser::new

jyn514:
This made things much more difficult and inconsistent, for no real benefit.

The major change is that the initial location is now a dummy location,
instead of being a real location that could be printed. Since the very
initial location will always be overwritten by the first token seen, this
should not cause any change in behavior.

Additionally, `Location::default` is now a public method. This is not ideal, but shouldn't hurt anything.

This makes #475 simpler, since the parser can now treat the `tokens` as the only source of tokens.

### DIFF
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -719,7 +719,6 @@ impl<'a> PreProcessor<'a> {
                 location,
             ));
         }
-        // TODO: remove(0) is bad and I should feel bad
         // TODO: this only returns the first error because anything else requires a refactor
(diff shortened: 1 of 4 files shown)


src/lex/cpp.rs
```diff
@@ -735,9 +721,8 @@ impl<'a> PreProcessor<'a> {
         }
         // TODO: remove(0) is bad and I should feel bad
```
jyn514, line 722:
```suggestion
```

PR #478, opened 2020-06-13
Get rid of unput in the lexer

jyn514:
- Move string concatenation to the parser
- Use std::iter::peekable to see if the next token is a string or not
- Don't consume the leading `"` of a string token in `parse_string`
- Get rid of `unput()` in the lexer

r? @hdamron17

### DIFF
--- a/src/parse/mod.rs
+++ b/src/parse/mod.rs
@@ -13,16 +13,9 @@ use crate::RecursionGuard;
 type Lexeme = CompileResult<Locatable<Token>>;
 type SyntaxResult<T> = Result<T, Locatable<SyntaxError>>;
 
-/// An iterator over `Lexeme`, but with a little more flexibility
-pub trait Lexer {
-    fn next(&mut self) -> Option<Lexeme>;
-}
-
-impl<I: Iterator<Item = Result<Locatable<Token>, E>>, E: Into<CompileError>> Lexer for I {
-    fn next(&mut self) -> Option<Lexeme> {
-        Iterator::next(self).map(|res| res.map_err(Into::into))
-    }
-}
+/// Trait alias for Iterator, but stable
+pub trait Lexer: Iterator<Item = Lexeme> {}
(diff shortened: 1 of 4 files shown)


src/parse/mod.rs
```diff
@@ -144,12 +137,52 @@ impl<I: Lexer> Parser<I> {
     }
     // don't use this, use next_token instead
     fn __impl_next_token(&mut self) -> Option<Locatable<Token>> {
+        // needed for string concatenation
+        assert!(self.current.is_none() || self.next.is_none());
```
jyn514, line 141: This is actually no longer needed.
```suggestion
```

src/parse/mod.rs
```diff
@@ -150,6 +143,44 @@ impl<I: Lexer> Parser<I> {
...
+                        match self.tokens.peek() {
+                            Some(Ok(Locatable {
+                                data: Token::Literal(Literal::Str(_)),
+                                location: next_location,
+                            })) => {
+                                location = location.merge(next_location);
+                                let mut s = match self.tokens.next().unwrap().unwrap().data {
+                                    Token::Literal(Literal::Str(s)) => s,
+                                    _ => unreachable!(),
+                                };
```
jyn514, line 161: Once `Token` is Copy, I can remove this in favor of just dereferencing the borrowed token.

jyn514:
> Why does parse_string have to start after the starting quote?

The starting quote has already been consumed by `next_char`, this is the reason I needed `unput` in the first place: https://github.com/jyn514/saltwater/blob/edc3d89a9be1b017ba5326c1bf1eed058a341526/src/lex/mod.rs#L717. Redoing that requires rewriting all of that function to use `peek()` instead of `next_char()`, which is at least a solid 2 hours of tedious coding.

> make sure the string location starts before the quotation mark

Yes, this is taken care of by https://github.com/jyn514/saltwater/blob/edc3d89a9be1b017ba5326c1bf1eed058a341526/src/lex/mod.rs#L718. I could use more tests for spans, though.

PR #484, opened 2020-07-06
Document that exponents aren't required for hex floats

jyn514:
r? @hdamron17 

Closes https://github.com/jyn514/saltwater/issues/483


hdamron17:
Why is the branch called apple-a-day..?

jyn514:
It keeps the docs away.

PR #486, opened 2020-07-20
Fix clippy warnings

jyn514:
This fixes warnings as far forward as 1.47, so hopefully this won't
fail CI again for a while.

### DIFF
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -1413,8 +1413,7 @@ pub(crate) mod test {
         assert_eq!(decl(left).unwrap().to_string(), right);
     }
     fn assert_extern_decl_display(s: &str) {
-        // TODO: this `auto` is such a hack
-        assert_decl_display(s, &s.to_string());
+        assert_decl_display(s, s);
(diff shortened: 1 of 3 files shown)


src/analyze/mod.rs
```diff
@@ -1417,7 +1414,7 @@ pub(crate) mod test {
     }
     fn assert_extern_decl_display(s: &str) {
         // TODO: this `auto` is such a hack
```
jyn514, line 1416: Outdated comment.
```suggestion
```

src/analyze/mod.rs
```diff
@@ -1417,7 +1414,7 @@ pub(crate) mod test {
...
-        assert_decl_display(s, &format!("{}", s));
+        assert_decl_display(s, &s.to_string());
```
jyn514, line 1417: This whole part of the code is a mess.
```suggestion
        assert_decl_display(s, s);
```

jyn514:
Merging despite the proptest failures because they also appear on master 🤦

PR #487, opened 2020-07-24
Add stddef.h

jyn514:
Closes https://github.com/jyn514/saltwater/issues/388. Gives a reasonable way to add additional header files.

Saltwater can, for the first time after 14 months, compile hello world with a GNU userland!

r? @pythondude325

### DIFF
--- a/src/analyze/expr.rs
+++ b/src/analyze/expr.rs
@@ -910,6 +910,7 @@ pub(super) fn literal(literal: Literal, location: Location) -> Expr {
     }
 }
 
+// 6.5.15 - Conditional operator
 fn pointer_promote(left: &mut Expr, right: &mut Expr) -> bool {
     let is_convertible_to_any_pointer = |expr: &Expr| {
         expr.ctype.is_void_pointer() || expr.ctype.is_char_pointer() || expr.is_null()
--- a/src/analyze/expr.rs
+++ b/src/analyze/expr.rs
@@ -911,12 +911,15 @@ pub(super) fn literal(literal: Literal, location: Location) -> Expr {
 }
 
 fn pointer_promote(left: &mut Expr, right: &mut Expr) -> bool {
+    let is_convertible_to_any_pointer = |expr: &Expr| {
+        expr.ctype.is_void_pointer() || expr.ctype.is_char_pointer() || expr.is_null()
+    };
     if left.ctype == right.ctype {
         true
-    } else if left.ctype.is_void_pointer() || left.ctype.is_char_pointer() || left.is_null() {
+    } else if is_convertible_to_any_pointer(left) && right.ctype.is_pointer() {
         left.ctype = right.ctype.clone();
         true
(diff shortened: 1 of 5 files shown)


src/analyze/expr.rs
```diff
@@ -911,12 +911,15 @@ pub(super) fn literal(literal: Literal, location: Location) -> Expr {
...
+        expr.ctype.is_void_pointer() || expr.ctype.is_char_pointer() || expr.is_null()
+    };
     if left.ctype == right.ctype {
         true
-    } else if left.ctype.is_void_pointer() || left.ctype.is_char_pointer() || left.is_null() {
+    } else if is_convertible_to_any_pointer(left) && right.ctype.is_pointer() {
         left.ctype = right.ctype.clone();
         true
-    } else if right.ctype.is_void_pointer() || right.ctype.is_char_pointer() || right.is_null() {
+    } else if is_convertible_to_any_pointer(right) && left.ctype.is_pointer() {
```
pythondude325, line 922: I don't know what these do. Can I get a link to the relevant spec section?

src/lex/cpp.rs
```diff
@@ -1066,15 +1066,33 @@ impl<'a> PreProcessor<'a> {
...
+                    .into();
+                (path, src)
+            }
+            Err(not_found) => {
+                let filename = match filename.file_name().and_then(|f| f.to_str()) {
+                    None => return Err(not_found),
+                    Some(f) => f,
+                };
+                match get_builtin_header(filename) {
+                    Some(file) => {
```
pythondude325, line 1085: Can you really override stddef.h? If so that's disgusting.

jyn514: My reasoning here is that if the standard library provides stddef.h, that's probably more accurate than the one I hacked together. The standard doesn't treat the compiler and library separately so it's up to us which we want to do.

src/analyze/expr.rs
jyn514: http://port70.net/~nsz/c/c11/n1570.html#6.5.15

jyn514: I also added a comment at the start of the function.

hdamron17: I feel like this should have been a separate pull request but oh well.

jyn514: Originally I had `#define va_arg(ap, type) (0 ? "va_arg is not currently supported by saltwater" : 0.0)`. But it turns out that I don't give _any_ sort of reasonable errors from macro expansion, so it didn't really matter.

```
stdarg_example.c:4:5 error: invalid program: incompatible types in ternary expression: 'char *' cannot be converted to 'double'
    va_start(args, n);
    ^^^^^^^^
```

jyn514: I improved the error messages a bit in the latest commit:

```c
stddef_example.c:5:21 error: invalid syntax: expected ')', got 'not_currently_supported'
    printf("%ul\n", offsetof(struct s { int i; }, i));
                    ^^^^^^^^
```

Still not great but better than before.

PR #488, opened 2020-07-24
[WIP] Add proper Repl

### LINKED ISSUE #372: Use rustyline for REPL

Stupremee:
This PR will introduce a proper Repl using [rustyline](https://crates.io/crates/rustyline) and the `jit` backend.

Only merge if #493 is merged
Resolves #372


src/bin/repl/helper.rs
```diff
@@ -84,5 +87,27 @@ impl Hinter for ReplHelper {
 impl Completer for ReplHelper {
     type Candidate = String;
```
jyn514, line 88: Now that the commands are `&'static str`, this could be as well.
```suggestion
    type Candidate = &'static str;
```

src/bin/repl/helper.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl Highlighter for ReplHelper {
+    fn highlight<'l>(&self, line: &'l str, pos: usize) -> Cow<'l, str> {
+        // TODO: Syntax highlighting.
+        self.highlighter.highlight(line, pos)
+    }
+
+    fn highlight_hint<'h>(&self, hint: &'h str) -> Cow<'h, str> {
+        let hint = Style::new().dimmed().paint(hint);
+        Cow::Owned(hint.to_string())
```
jyn514, line 37: This could be `Cow::Borrowed` I think.
```suggestion
        Cow::Borrowed(hint)
```

src/bin/repl/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+            .history_ignore_dups(false)
+            .completion_type(CompletionType::List)
+            .edit_mode(EditMode::Emacs)
+            .max_history_size(1000)
+            .tab_stop(4)
+            .build();
+        let mut editor = Editor::with_config(config);
+
+        let commands = default_commands();
+        let helper = ReplHelper::new(commands.keys().map(|x| *x).collect());
```
jyn514, line 37:
```suggestion
        let helper = ReplHelper::new(commands.keys().copied().collect());
```

src/bin/repl/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+        Some(history)
+    }
+
+    fn process_line(&mut self, line: String) {
+        self.editor.add_history_entry(line.clone());
+
+        if line.trim().starts_with(PREFIX) {
+            let name = line.split(' ').next().unwrap();
+
+            match self.commands.get(&name[1..]) {
```
jyn514, line 90: Will this panic on `:`?

src/bin/repl/commands.rs
```diff
@@ -0,0 +1,26 @@
...
+    map
+}
+
+fn help_command(_repl: &mut Repl, _args: &str) {
+    println!(
+        "
+Available commands:
+    {p}help      Shows this message
+    {p}quit|q    Quits the repl
+",
```
jyn514, line 18: This prints an extra newline at the start and end.
```suggestion
    print!(
        "\
Available commands:
    {p}help      Shows this message
    {p}quit|q    Quits the repl
",
```

src/bin/repl/helper.rs
```diff
@@ -0,0 +1,113 @@
...
+        let mut stack = vec![];
+
+        for c in input.chars() {
+            match c {
+                '(' | '[' | '{' => stack.push(c),
+                ')' | ']' | '}' => match (stack.pop(), c) {
+                    (Some('('), ')') | (Some('['), ']') | (Some('{'), '}') => {}
+                    (_, _) => {
+                        return Ok(ValidationResult::Invalid(Some(
+                            "unclosed delimiter".to_string(),
```
jyn514, line 57: This is `a)`, not `(a`.
```suggestion
                            "extra closing delimiter".to_string(),
```

src/bin/repl/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    fn load_history(&mut self) -> Option<()> {
+        let path = Self::history_path()?;
+        self.editor.load_history(&path).ok()
+    }
+
+    fn history_path() -> Option<PathBuf> {
+        let mut history = home_dir()?;
+        history.push(".saltwater_history");
```
jyn514, line 80: This should use [`data_dir`](https://docs.rs/dirs-next/1.0.1/dirs_next/fn.data_dir.html) or `cache_dir` I think.
```suggestion
        let mut history = data_dir()?;
        history.push("saltwater_history");
```

src/bin/swcci.rs
```diff
@@ -0,0 +1,12 @@
+mod repl;
+
+fn main() {
+    let mut repl = repl::Repl::new();
+    match repl.run() {
+        Ok(_) => {}
+        Err(err) => {
+            println!("Unknown error in the REPL occurred: {}", err);
```
jyn514, line 8: This seems a little verbose:
```suggestion
            println!("error: {}", err);
```

src/bin/repl/mod.rs
Stupremee: No, it will print `unknown command ':'`.

src/bin/repl/helper.rs
Stupremee: Unfortunately, it can't.

jyn514: Hmm I see, and you can't change the return type since it's a trait. Ok, not worth worrying about then.

src/bin/repl/mod.rs
```diff
@@ -93,7 +98,92 @@ impl Repl {
...
+        let mut jit = JIT::from(module);
+        jit.finalize();
+        let execute_fun = jit
+            .get_compiled_function("execute")
+            .expect("this is not good.");
+
+        match expr_ty {
+            Type::Long(signed) => {
+                let result = unsafe {
+                    let execute: unsafe extern "C" fn() -> u64 = std::mem::transmute(execute_fun);
```
jyn514, line 128: Any reason you can't return `i64` directly here? They're represented the same at the hardware level.
```suggestion
                    let execute: unsafe extern "C" fn() -> i64 = std::mem::transmute(execute_fun);
```

src/bin/repl/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+fn wrap_expr(expr: Expr) -> Locatable<Declaration> {
+    let fun = hir::Variable {
+        ctype: types::Type::Function(types::FunctionType {
+            return_type: Box::new(expr.ctype.clone()),
+            params: vec![],
+            varargs: false,
+        }),
+        storage_class: data::StorageClass::Extern,
+        qualifiers: Default::default(),
+        id: saltwater::InternedStr::get_or_intern("execute"),
```
jyn514, line 159:
```suggestion
        id: "execute".into(),
```

src/bin/repl/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+    };
+
+    let span = expr.location;
+    let return_stmt = span.with(hir::StmtType::Return(Some(expr)));
+    let init = hir::Initializer::FunctionBody(vec![return_stmt]);
+    let decl = hir::Declaration {
+        // FIXME: Currently doesn't work. Just make insert() pub?
+        symbol: fun.insert(),
```
jyn514, line 167: Hmm, the reason I make this private was because I was thinking about parsing and didn't want people to accidentally add a new identifier ... but it does seem like that would be a major use case for any programmatic use. I'm fine with making it public :)

src/bin/repl/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+    let expr = parser.expr()?;
+
+    let mut analyzer = PureAnalyzer::new();
+    let expr = analyzer.expr(expr);
+
+    // FIXME: error_handler is private so this doesn't work right now.
+    // Please review and propose a solution.
+    // if let Some(err) = analyzer.error_handler.pop_front() {
+    //     return Err(err);
+    // }
```
jyn514, line 186: Oops, there isn't actually a way to get errors from `PureAnalyzer`.

@pythondude325 I think we could add an `errors()` method here? Where it just returns `mem::take(&mut self.error_handler)`?

src/bin/swcc.rs
```diff
@@ -554,7 +554,7 @@ mod backtrace {
 
 #[cfg(feature = "salty")]
 fn play_scream() -> Result<(), ()> {
-    const SCREAM: &[u8] = include_bytes!("data/R2D2-Scream.ogg");
+    const SCREAM: &[u8] = include_bytes!("../data/R2D2-Scream.ogg");
```
jyn514, line 557:
```suggestion
    const SCREAM: &[u8] = include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "data/R2D2-Scream.ogg")));
```

src/bin/repl/helper.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+impl Hinter for ReplHelper {
+    fn hint(&self, line: &str, pos: usize, _ctx: &Context<'_>) -> Option<String> {
+        let start = &line[..pos];
+        if !start.starts_with(PREFIX) {
+            return None;
+        }
+        let start = &start[1..];
```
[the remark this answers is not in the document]
Stupremee: I made the `PREFIX` a `char` to avoid stuff like that

src/bin/repl/mod.rs
Stupremee: Yes I can do that, but then I still have to cast it to a `u64` if it's unsigned, so there's not really a difference to the current code

src/bin/repl/mod.rs
```diff
@@ -21,6 +22,15 @@ const PREFIX: char = ':';
...
 const PROMPT: &str = ">> ";
 
+macro_rules! execute {
+    ($fun:ident, $ty:path, $action:expr) => {
+        $action(unsafe {
+            let execute: unsafe extern "C" fn() -> $ty = std::mem::transmute($fun);
+            execute()
+        });
+    };
+}
```
jyn514, line 32: This is used like
```rust
execute!(fun, bool, |b| println!("=> {}", b)),
```
I think it would be easier and simpler to do `println!("=> {}", execute!(fun, bool))`, which means you don't need `$action` or to create a new closure.

src/bin/repl/mod.rs
```diff
@@ -114,26 +124,37 @@ impl Repl {
...
+                false => println!("=> {}", x as u32),
+            }),
+            Type::Long(signed) => execute!(fun, i64, |x| match signed {
+                true => println!("=> {}", x),
+                false => println!("=> {}", x as u64),
+            }),
+            Type::Float => execute!(fun, f32, |f| println!("=> {}", f)),
+            Type::Double => execute!(fun, f64, |f| println!("=> {}", f)),
+
+            Type::Char(_) => execute!(fun, char, |c| println!("=> {}", c)),
```
jyn514, line 151: This doesn't look quite right ... in Rust, `char` is a [unicode scalar character](https://doc.rust-lang.org/std/primitive.char.html) which is represented as 4 bytes. In C, it's an arbitrary byte, which could be invalid utf8 (`0xff`).
```suggestion
            Type::Char(_) => execute!(fun, u8, |c| println!("=> {}", if c.is_ascii() { c as char } else { `hex::encode(&[c]) })),
```
https://docs.rs/hex/0.4.2/hex/fn.encode.html

src/bin/repl/mod.rs
```diff
@@ -114,26 +124,37 @@ impl Repl {
...
+            Type::Bool => execute!(fun, bool, |b| println!("=> {}", b)),
+            Type::Void => unsafe {
+                let execute: unsafe extern "C" fn() = std::mem::transmute(fun);
+                execute()
+            },
```
jyn514, line 156:
```suggestion
            Type::Void => execute!(fun, (), |_| {}),
```

Stupremee:
I have to reopen because I accidentally deleted my fork (:sweat_smile:)

Issue #449, opened 2020-05-23
Cannot use function macros in #if macros


jyn514:
The following code has no output when run with `-E`:
```c
#define f(a) 1
#if f(a)
success
#endif
```
The issue is that I call `replace` on each token individually, not considering the tokens following. Relevant code: https://github.com/jyn514/rcc/blob/12a71429afd9f7c5d2576f509527c8e91dba81e5/src/lex/cpp.rs#L701

The fix is to pass in the following tokens as well as the current token.

jyn514:
Introduced by #441

PR #492, opened 2020-07-29
Close #453. Builtin macros such as `__LINE__` (not function macros like `__has_header()`)

### LINKED ISSUE #453: Missing predefined macros
[6.10.8.1 Mandatory macros](http://port70.net/~nsz/c/c11/n1570.html#6.10.8.1)

- [x] `__LINE__` - tricky, can't be passed through to `replace()` literally. This might be as simple as updating `self.definitions` whenever someone calls `line()`?
- [ ] `__COLUMN__` - tricky since we don't keep track of this explicitly and recalculate it. Probably needs yet more preprocessing before passing things through to `replace()`.
- [x] `__FILE__` - easy, but might need a refactor to put `PreProcessor` and `FileProcessor` back together.
- [x] `__DATE__` - easy, but needs a date-time dependency
- [x] `__TIME__` - same as `__DATE__`

[6.10.8.3 Conditional feature macros](http://port70.net/~nsz/c/c11/n1570.html#6.10.8.3)

- [ ] ` __STDC_IEC_559__ ` - trivial, good first issue. Code should go near https://github.com/jyn514/rcc/blob/180b27bb90bfcc02e2d4cba3afdd0a4dedccd662/src/lex/cpp.rs#L330.

Note that `__LINE__` should be the line at time of replacement, not within a definition. For example:
```c
$ clang -x c - -E -P
#define a __LINE__
a

2
```

hdamron17:
This will provide the functionality to define builtin macros. Those in #453 and `__has_header()` will be implemented as proof of concept.

### DIFF
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -1169,8 +1169,8 @@ impl<'a> PreProcessor<'a> {
 fn int_def(i: i32) -> Definition {
     Definition::Object(vec![LiteralToken::Int(RcStr::from(i.to_string())).into()])
 }
-fn str_def<S: ToString>(s: S) -> Definition {
-    let rcstr = RcStr::from(format!("\"{}\"", s.to_string().replace(r#"""#, r#"\""#)));
+fn str_def<S: Into<String>>(s: S) -> Definition {
+    let rcstr = RcStr::from(format!("\"{}\"", s.into().replace(r#"""#, r#"\""#)));
     Definition::Object(vec![LiteralToken::Str(vec![rcstr]).into()])
 }
 
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -280,6 +280,7 @@ impl<'a> PreProcessor<'a> {
             match token {
                 PendingToken::Replaced(t) => Some(Ok(Locatable::new(t, location))),
                 PendingToken::NeedsReplacement(token) => {
+                    self.update_builtin_definitions();
                     let mut replacement_list =
                         replace(&self.definitions, token, &mut self.file_processor, location)
                             .into_iter();
@@ -342,6 +343,7 @@ impl<'a> PreProcessor<'a> {
             "__STDC_NO_COMPLEX__".into() => int(1),
             "__STDC_NO_THREADS__".into() => int(1),
             "__STDC_NO_VLA__".into() => int(1),
+            "__STDC_IEC_559__".into() => int(1),
         };
         definitions.extend(user_definitions);
         let mut search_path = vec![
@@ -354,14 +356,16 @@ impl<'a> PreProcessor<'a> {
 
         let file_processor = FileProcessor::new(chars, filename, debug);
 
-        Self {
+        let mut new_cpp = Self {
             error_handler: Default::default(),
             nested_ifs: Default::default(),
             pending: Default::default(),
             search_path,
             definitions,
             file_processor,
-        }
+        };
+        new_cpp.update_builtin_definitions(); // So they are defined from the start
+        new_cpp
     }
 
     /// Return all warnings found so far.
@@ -590,6 +594,7 @@ impl<'a> PreProcessor<'a> {
         let location = self.span(start);
 
         // TODO: is this unwrap safe? there should only be scalar types in a cpp directive...
+        self.update_builtin_definitions();
         match Self::cpp_expr(&self.definitions, lex_tokens.into_iter(), location)?
             .truthy(&mut self.error_handler)
             .constexpr()?
@@ -973,6 +978,7 @@ impl<'a> PreProcessor<'a> {
                     ))
                 }
             };
+            self.update_builtin_definitions();
             match replace(
                 &self.definitions,
                 Token::Id(id),
@@ -1151,6 +1157,25 @@ impl<'a> PreProcessor<'a> {
             return Err(self.span(start).error(error));
         }
         Ok(ret)
+    }
+
+    fn update_builtin_definitions(&mut self) {
+        use Definition::Object;
+        let int_def = |i: i32| Object(vec![LiteralToken::Int(RcStr::from(i.to_string())).into()]);
+        let str_def = |s: &str| {
+            let rcstr = RcStr::from(format!("\"{}\"", s));
+            Object(vec![LiteralToken::Str(vec![rcstr]).into()])
+        };
+        self.definitions.extend(map! {
+            "__LINE__".into() => int_def((self.line() + 1) as i32),
+            "__FILE__".into() => str_def(self.file_processor.path().to_str().unwrap()),  // Fails if non-Unicode
+            "__DATE__".into() => {
+                str_def(&time::OffsetDateTime::now_local().format("%b %_d %Y"))
+            },
+            "__TIME__".into() => {
+                str_def(&time::OffsetDateTime::now_local().format("%H:%M:%S"))
+            },
+        })
     }
 }
 
@@ -1831,5 +1856,52 @@ h",
         );
         assert!(cpp("#define f(x) #y\nf(0)").any(|x| x.is_err()));
         assert!(cpp("#define f(x) #+\nf(0)").any(|x| x.is_err()));
+    }
+
+    #[test]
+    fn builtins_line() {
+        assert_same("__LINE__", "1");
+        assert_same("\n\n\n\n\n\n\n\n\n__LINE__", "10");
+        assert_same(
+            "#ifdef __LINE__
+            1
+            #endif",
+            "1",
+        );
+        assert_same(
+            "
+            
+            #if __LINE__ == 3
+            1
+            #endif",
+            "1",
+        );
+        assert_same(
+            "#define LINE __LINE__
+            
+
+            LINE",
+            "4",
+        );
+    }
+    #[test]
+    fn builtins_file() {
+        let filename = "helloworld.c";
+        let mut cpp = PreProcessorBuilder::new("__FILE__")
+            .filename(filename)
+            .build();
+        let token = cpp.next_non_whitespace().unwrap().unwrap().data;
+        if let Token::Literal(LiteralToken::Str(rcstrs)) = token {
+            assert_eq!(
+                rcstrs.first().unwrap().as_str(),
+                format!("\"{}\"", filename)
+            );
+        } else {
+            panic!();
+        }
+    }
+    #[test]
+    fn builtins_date_time() {
+        todo!();
     }
 }
--- a/src/lex/cpp.rs
+++ b/src/lex/cpp.rs
@@ -326,24 +326,23 @@ impl<'a> PreProcessor<'a> {
             "{}-{}-{}",
             TARGET.architecture, TARGET.operating_system, TARGET.environment
         );
-        let int = |i: i64| {
-            Definition::Object(vec![Token::Literal(LiteralToken::Int(RcStr::from(
-                i.to_string(),
-            )))])
-        };
+
+        let now = time::OffsetDateTime::now_local();
 
         #[allow(clippy::inconsistent_digit_grouping)]
         let mut definitions = map! {
-            format!("__{}__", TARGET.architecture).into() => int(1),
-            format!("__{}__", TARGET.operating_system).into() => int(1),
-            "__STDC__".into() => int(1),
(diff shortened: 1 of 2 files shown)


hdamron17:
The `__has_header` macro is only usable in a `#if` expression so I think I'll need to fix #449 first.

jyn514:
@hdamron17 interested in picking this back up now that #449 is fixed?

hdamron17:
> @hdamron17 interested in picking this back up now that #449 is fixed?

I'm still thinking through how everything will tie together, especially the function macros (others are easy and I could switch this to only do those if you want).

jyn514:
Let's do only object macros for now yeah. That will hit the vast majority of use cases.

src/lex/cpp.rs
```diff
@@ -1832,4 +1857,51 @@ h",
...
+                rcstrs.first().unwrap().as_str(),
+                format!("\"{}\"", filename)
+            );
+        } else {
+            panic!();
+        }
+    }
+    #[test]
+    fn builtins_date_time() {
+        todo!();
```
hdamron17, line 1905: I have no clue how to test `__DATE__` and `__TIME__` since they depend on current time.

src/lex/cpp.rs
```diff
@@ -354,14 +356,16 @@ impl<'a> PreProcessor<'a> {
...
+        let mut new_cpp = Self {
             error_handler: Default::default(),
             nested_ifs: Default::default(),
             pending: Default::default(),
             search_path,
             definitions,
             file_processor,
-        }
+        };
+        new_cpp.update_builtin_definitions(); // So they are defined from the start
```
jyn514, line 367: Is there a reason this needs to be done eagerly instead of lazily? I think if you only call it just before any call to replace you should be fine.

src/lex/cpp.rs
```diff
@@ -590,6 +594,7 @@ impl<'a> PreProcessor<'a> {
         let location = self.span(start);
 
         // TODO: is this unwrap safe? there should only be scalar types in a cpp directive...
+        self.update_builtin_definitions();
```
jyn514, line 597: Same comment, I think this only needs to be done just before `replace()`.

src/lex/cpp.rs
```diff
@@ -1152,6 +1158,25 @@ impl<'a> PreProcessor<'a> {
...
         Ok(ret)
     }
+
+    fn update_builtin_definitions(&mut self) {
+        use Definition::Object;
+        let int_def = |i: i32| Object(vec![LiteralToken::Int(RcStr::from(i.to_string())).into()]);
+        let str_def = |s: &str| {
+            let rcstr = RcStr::from(format!("\"{}\"", s));
+            Object(vec![LiteralToken::Str(vec![rcstr]).into()])
+        };
```
jyn514, line 1168: I feel like I see these a lot, maybe we could make them functions so they can be reused?

src/lex/cpp.rs
```diff
@@ -1152,6 +1158,25 @@ impl<'a> PreProcessor<'a> {
...
+        self.definitions.extend(map! {
+            "__LINE__".into() => int_def((self.line() + 1) as i32),
+            "__FILE__".into() => str_def(self.file_processor.path().to_str().unwrap()),  // Fails if non-Unicode
```
jyn514, line 1171: Non Unicode file paths are recoverable I think. This defines it to be empty if it isn't valid UTF-8. Another alternative is to use `to_str_lossy`, but that might not be supported by RcStr.
```suggestion
            "__FILE__".into() => str_def(self.file_processor.path().to_str().unwrap_or_default()),  // Fails if non-Unicode
```

src/lex/cpp.rs
```diff
@@ -1832,4 +1857,51 @@ h",
...
+            #if __LINE__ == 3
+            1
+            #endif",
+            "1",
+        );
+        assert_same(
+            "#define LINE __LINE__
+            
+
+            LINE",
```
jyn514, line 1883: This is evil, I love it :laughing: can you confirm that clang does the same?

src/lex/cpp.rs
jyn514: You could create a `OffsetDateTime` the same way you do in the code and make sure the difference between that and the macro one is less than a few seconds.

src/lex/cpp.rs
```diff
@@ -1152,6 +1158,25 @@ impl<'a> PreProcessor<'a> {
...
+        };
+        self.definitions.extend(map! {
+            "__LINE__".into() => int_def((self.line() + 1) as i32),
+            "__FILE__".into() => str_def(self.file_processor.path().to_str().unwrap()),  // Fails if non-Unicode
+            "__DATE__".into() => {
+                str_def(&time::OffsetDateTime::now_local().format("%b %_d %Y"))
+            },
+            "__TIME__".into() => {
+                str_def(&time::OffsetDateTime::now_local().format("%H:%M:%S"))
+            },
```
jyn514, line 1177: The standard says DATE and TIME should remain the same throughout compilation: http://port70.net/~nsz/c/c11/n1570.html#6.10.8. So this should be moved to `Preprocessor::new` I think.

src/lex/cpp.rs
```diff
@@ -1832,4 +1857,51 @@ h",
...
+            "1",
+        );
+        assert_same(
+            "
+            
+            #if __LINE__ == 3
+            1
+            #endif",
+            "1",
+        );
```
jyn514, line 1878: Can you add a test case where LINE is used twice to make sure the old definition gets overridden?

src/lex/cpp.rs
hdamron17: We do it once to ensure the builtins are considered defined for `#ifdef` which doesn't do any replacement.

src/lex/cpp.rs
hdamron17: I put it here because `cpp_expr` is a free function so I can't put it in front of the internal `replace`. However, `cpp_expr` only looks at one line so nothing should change. I should put it before the TODO comment though.

src/lex/cpp.rs
hdamron17: I'll leave that for another PR since we'd want to change it in all of the current places as well. This is a little bit different and more complex than usual because it wraps in `Definition::Object(vec![...])`.

src/lex/cpp.rs
hdamron17: Anything other than panicking would give the wrong behaviour, no? I can do `_or_default()` if you'd like though (I don't think `RcStr` will take lossy).

src/lex/cpp.rs
hdamron17: Ah, didn't see that. I'll change it. This also makes testing `__DATE__` and `__TIME__` much easier because I can mock the preprocessor to have certain times to ensure the formatting is correct.

src/lex/cpp.rs
hdamron17: Confirmed

src/lex/cpp.rs
```diff
@@ -342,6 +343,7 @@ impl<'a> PreProcessor<'a> {
             "__STDC_NO_COMPLEX__".into() => int(1),
             "__STDC_NO_THREADS__".into() => int(1),
             "__STDC_NO_VLA__".into() => int(1),
+            "__STDC_IEC_559__".into() => int(1),
```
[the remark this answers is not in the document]
hdamron17: Wait so where should I get the value? I would need to read some of the standard library to find the value somewhere? Also, the `__GCC_IEC_559` is a gcc thing so I'm not sure what else I'd find in the std library.

hdamron17:
The standard says "None of these macro names, nor the identifier defined, shall be the subject of a #define or a #undef preprocessing directive" ([6.10.8p2](http://port70.net/~nsz/c/c11/n1570.html#6.10.8p2)). In clang, it is a warning and then successfully redefines. For us, any of the constantly redefined macros (like `__LINE__`), this would override any user redefinition. Is this something we care about since the standard says no?

src/lex/cpp.rs
hdamron17: Never mind, I went ahead and changed it.

jyn514:
'shall not' means this is a constraint on the user, not on the compiler, so technically the behavior is undefined. It would be friendly to give a warning but I'm fine having it do unexpected things since this is a really bad idea anyway.

hdamron17:
I think it would be hard to maintain giving a warning since we don't differentiate the builtins anywhere. I'll leave it such that __LINE__ and __FILE__ cannot be replaced but the others can (although none of them *should*).

src/lex/cpp.rs
hdamron17: Done

hdamron17:
Interestingly, clang doesn't define `__STDC_IEC_559__`.

hdamron17:
Also, according to [GCC](https://gcc.gnu.org/c99status.html), "GCC does not define `__STDC_IEC_559__` or implement the associated standard pragmas"

hdamron17:
Even though GCC says it doesn't, `__STDC_IEC_559__` is defined as 1 on my Ubuntu machine. On Mac with clang, nothing.

src/lex/cpp.rs
jyn514: Hmm, can you make it an error instead of a panic then? I don't think we should panic on user errors except in extreme cases (e.g. https://github.com/jyn514/saltwater/blob/master/src/ir/mod.rs#L73). See https://github.com/jyn514/saltwater/blob/0e36d97ac31a1875d6a1276b1f8840fedcef65e8/src/lex/cpp.rs#L442 for an example.

src/lex/cpp.rs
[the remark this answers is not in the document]
jyn514: Right, I'm saying don't set it at all, since we can't guarentee it.

src/lex/cpp.rs
hdamron17: I can do `_lossy` but I'm not sure how that works since I'm `format!`ing it into quotation marks before it goes to `RcStr` so I'm not sure if this does what you want. Also, can paths have quotation marks in them?

hdamron17:
Just waiting on `__STDC_IEC_559__` now.

r? @jyn514

src/lex/cpp.rs
```diff
@@ -1161,16 +1161,16 @@ impl<'a> PreProcessor<'a> {
...
     }
 }
 
 fn int_def(i: i32) -> Definition {
     Definition::Object(vec![LiteralToken::Int(RcStr::from(i.to_string())).into()])
 }
-fn str_def(s: &str) -> Definition {
-    let rcstr = RcStr::from(format!("\"{}\"", s));
+fn str_def<S: ToString>(s: S) -> Definition {
+    let rcstr = RcStr::from(format!("\"{}\"", s.to_string().replace(r#"""#, r#"\""#)));
```
jyn514, line 1173: Use `Into` instead; `ToString` is meant for displaying things to the end user and is more flexible than we need here (it includes things like numbers and ip addresses): https://doc.rust-lang.org/std/string/trait.ToString.html

```suggestion
fn str_def<S: Into<String>>(s: S) -> Definition {
    let rcstr = RcStr::from(format!("\"{}\"", s.into().replace(r#"""#, r#"\""#)));
```

hdamron17: Good to know, thanks!

src/lex/cpp.rs
```diff
@@ -325,23 +326,23 @@ impl<'a> PreProcessor<'a> {
...
+            format!("__{}__", TARGET.architecture).into() => int_def(1),
+            format!("__{}__", TARGET.operating_system).into() => int_def(1),
+            "__STDC__".into() => int_def(1),
+            "__STDC_HOSTED__".into() => int_def(1),
+            "__STDC_VERSION__".into() => int_def(2011_12),
+            "__STDC_NO_ATOMICS__".into() => int_def(1),
+            "__STDC_NO_COMPLEX__".into() => int_def(1),
+            "__STDC_NO_THREADS__".into() => int_def(1),
+            "__STDC_NO_VLA__".into() => int_def(1),
+            "__STDC_IEC_559__".into() => int_def(1),
```
jyn514: Looks good to me once this is removed.
```suggestion
```

hdamron17: Done

Issue #20, opened 2019-08-28
Statements


jyn514:
```
int main() { int i = 5, j = 0; while (i >= 0) { j += 1; i -= 1; } }
thread 'main' panicked at 'not yet implemented: almost every statement', src/ir.rs:296:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

- [x] while loops in https://github.com/jyn514/rcc/commit/41d78767322a8e564484ae492ef2601a255a56a5
- [x] continue/break in https://github.com/jyn514/rcc/commit/41d78767322a8e564484ae492ef2601a255a56a5
- [x] do while loops in https://github.com/jyn514/rcc/commit/6cb4e507396ae1dec60404453c03f834e16adba4
- [x] for loops in https://github.com/jyn514/rcc/commit/883e12692574b2986231bf5082bc14dcb6303481
- [x] switch/case/default in https://github.com/jyn514/rcc/commit/4a71adaf28e700993982c70942f99ceaae658b97
- [ ] goto/label

jyn514:
This is finished for everything except `goto` for a label that occurs later in the code, moving to a new issue.

PR #497, opened 2020-08-09
Close #494. Update pico-args with short-space-opt feature and use str for paths

### LINKED ISSUE #494: Optional whitespace between command line argument key and value
This is blocked by [#19](https://github.com/RazrFalcon/pico-args/issues/19) on pico-args, but I wanted to make an issue here as well. It is common practice to leave off the space for `-I`, `-L`, and `-l` args, but currently we require a space. This will be needed for CMake compatibility as well, I think.

Expected
--------
`swcc -I/usr/local/include sample.c` successfully compiles sample.c using includes from `/usr/local/include`

Actual
------
The above command fails with message `swcc: error parsing args: unused arguments left: -I/usr/local/include`

hdamron17:
Close #494. Following pico-args [#20](https://github.com/RazrFalcon/pico-args/pull/20), this pull request takes advantage of the short-space-opt feature which makes `-Iinclude` possible. In order to do this, we had to use `str` instead of `OsStr` which isn't ideal.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ hexponent = "0.3"
 thiserror = "^1.0.10"
 target-lexicon = "0.10"
 tempfile = { version = "3", optional = true }
-pico-args = { version = "0.3", optional = true }
+pico-args = { version = "0.3.4", optional = true, features = ["short-space-opt"] }
 lasso = "0.2"
 codespan = { version = ">=0.9.5", default-features = false }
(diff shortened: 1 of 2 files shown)


jyn514:
This is making me realize we need tests for CLI arguments :/ I'll test manually so we don't have to block this PR but I think we should add tests at some point.

hdamron17:
I was thinking the same thing..

PR #502, opened 2020-08-11
Don't emit further type errors if already encountered type error

### LINKED ISSUE #458: Do not report type error when one type is "<type error>"
Since "\<type error\>" is a type generated only in the case of an already reported error, reporting it in a type error is redundant.

#### Code
```c
int main() {
  return ++i;
}
```

#### Expected Error (clang)
```
evil-test.c:2:12: error: use of undeclared identifier 'i'
  return ++i;
           ^
1 error generated.
```

#### Actual Error
```
evil-test.c:2:12 error: invalid program: use of undeclared identifier 'i'
  return ++i;
           ^
evil-test.c:2:10 error: invalid program: cannot assign to rvalue
  return ++i;
         ^^
evil-test.c:2:10 error: invalid program: cannot assign to rvalue
  return ++i;
         ^^
evil-test.c:2:10 error: invalid program: invalid operators for '+' (expected either arithmetic types or pointer operation, got '<type error> + <type error>'
  return ++i;
         ^^
4 errors generated
```

ryankeleti:
This fixes #458 by checking for `Type::Error` before reporting other errors. Also undeclared identifiers are now set to be `lval`s.

Not sure if my code is the most idiomatic / if I got all cases, but helped to get a better understanding of the analyzer.

### DIFF
--- a/src/analyze/expr.rs
+++ b/src/analyze/expr.rs
@@ -58,11 +58,19 @@ impl PureAnalyzer {
                 let struct_type = match &inner.ctype {
                     Type::Pointer(ctype, _) => match &**ctype {
                         Type::Union(_) | Type::Struct(_) => (**ctype).clone(),
+                        Type::Error => {
+                            // already type error
+                            return inner;
+                        }
                         other => {
                             self.err(SemanticError::NotAStruct(other.clone()), inner.location);
                             return inner;
                         }
                     },
+                    Type::Error => {
+                        // already type error
+                        return inner;
+                    }
                     other => {
                         self.err(SemanticError::NotAPointer(other.clone()), inner.location);
                         return inner;
@@ -82,6 +90,7 @@ impl PureAnalyzer {
                         let ctype = (**t).clone();
                         inner.indirection(true, ctype)
                     }
+                    Type::Error => inner, // already type error
                     _ => {
                         self.err(
                             SemanticError::NotAPointer(inner.ctype.clone()),
@@ -213,7 +222,10 @@ impl PureAnalyzer {
         };
         let location = left.location.merge(right.location);
         if let Some(ctype) = non_scalar {
-            self.err(SemanticError::NonIntegralExpr(ctype.clone()), location);
+            // if not already type error
+            if *ctype != Type::Error {
+                self.err(SemanticError::NonIntegralExpr(ctype.clone()), location);
+            }
         }
         let (promoted_expr, next) = Expr::binary_promote(left, right, &mut self.error_handler);
         Expr {
@@ -227,6 +239,7 @@ impl PureAnalyzer {
     fn parse_id(&mut self, name: InternedStr, location: Location) -> Expr {
         let mut pretend_zero = Expr::zero(location);
         pretend_zero.ctype = Type::Error;
+        pretend_zero.lval = true; // set undeclared identifier as lval
         match self.scope.get(&name) {
             None => {
                 self.err(SemanticError::UndeclaredVar(name), location);
@@ -283,7 +296,10 @@ impl PureAnalyzer {
         } else {
             let (left_expr, right_expr) = (left.rval(), right.rval());
             // p1 == p2
-            if !((left_expr.ctype.is_pointer() && left_expr.ctype == right_expr.ctype)
+            // first check that not already type error
+            if left_expr.ctype != Type::Error
+                && right_expr.ctype != Type::Error // Maybe I should pull this out of the if...
+                && !((left_expr.ctype.is_pointer() && left_expr.ctype == right_expr.ctype)
                 // equality operations have different rules :(
                 || ((token == ComparisonToken::EqualEqual || token == ComparisonToken::NotEqual)
                     // shoot me now
@@ -320,6 +336,11 @@ impl PureAnalyzer {
     // 6.5.5 Multiplicative operators
     fn mul(&mut self, left: Expr, right: Expr, op: BinaryOp) -> Expr {
         let location = left.location.merge(right.location);
+
+        // TODO: is this the right behavior?
+        if left.ctype == Type::Error || right.ctype == Type::Error {
+            return left;
+        }
 
         if op == BinaryOp::Mod && !(left.ctype.is_integral() && right.ctype.is_integral()) {
             self.err(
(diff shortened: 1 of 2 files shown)


src/analyze/expr.rs
```diff
@@ -58,11 +58,19 @@ impl PureAnalyzer {
                 let struct_type = match &inner.ctype {
                     Type::Pointer(ctype, _) => match &**ctype {
                         Type::Union(_) | Type::Struct(_) => (**ctype).clone(),
+                        Type::Error => {
+                            // already type error
+                            return inner;
+                        }
```
jyn514, line 64: I think this is pretty clear from the code.
```suggestion
                        Type::Error => return inner,
```

src/analyze/expr.rs
```diff
@@ -58,11 +58,19 @@ impl PureAnalyzer {
...
                         other => {
                             self.err(SemanticError::NotAStruct(other.clone()), inner.location);
                             return inner;
                         }
                     },
+                    Type::Error => {
+                        // already type error
+                        return inner;
+                    }
```
jyn514, line 73:
```suggestion
                    Type::Error => return inner,
```

src/analyze/expr.rs
```diff
@@ -82,6 +90,7 @@ impl PureAnalyzer {
                         let ctype = (**t).clone();
                         inner.indirection(true, ctype)
                     }
+                    Type::Error => inner, // already type error
```
jyn514, line 93:
```suggestion
                    Type::Error => inner,
```

src/analyze/expr.rs
```diff
@@ -321,6 +337,11 @@ impl PureAnalyzer {
     fn mul(&mut self, left: Expr, right: Expr, op: BinaryOp) -> Expr {
         let location = left.location.merge(right.location);
 
+        // TODO: is this the right behavior?
+        if left.ctype == Type::Error || right.ctype == Type::Error {
+            return left;
+        }
```
jyn514, line 343: If any error is emitted, `swcc` won't emit a binary, so the returned expression is just for error recovery. I do think we should return the expression that was an error though, to avoid cascading errors.
```suggestion
        if left.ctype == Type::Error {
            return left;
        } else if right.ctype == Type::Error {
            return right;
        }
```

jyn514:
```
testing tests/runner-tests/expr/explicit_casts/6.c

thread 'run_all' panicked at 'verification error: - inst1 (return v0): arg 0 (v0) has type i64, must match function signature of i32

note: while compiling function u0:0() -> i32 system_v {

block0:

    v0 = iconst.i64 5

    return v0

}
```
Looks like somewhere is returning the valid expression after an invalid operation instead returning `Type::Error`.

src/analyze/expr.rs
```diff
@@ -1270,10 +1315,18 @@ impl Expr {
...
+            // check not type error and that if rval is a pointer, it is not a type error
+            if *ctype != Type::Error {
+                if let Type::Pointer(a, _) = expr.ctype.clone() {
+                    if *a != Type::Error {
+                        error_handler.error(
+                            SemanticError::InvalidCast(expr.ctype.clone(), ctype.clone()),
+                            expr.location,
+                        );
+                    }
+                }
```
jyn514, line 1328: This looks like the incorrect logic, if it's not a pointer you should still report the error. Is there a reason you're special casing pointers here?

ryankeleti: This does seem incorrect, I don't remember exactly why I chose to do this. I think it got rid of a cascading error in one case, but the nested if hides an invalid cast for nonpointers like you say. For now I can just remove the pointer part

jyn514:
```
testing tests/runner-tests/decls/test_initializers/8.c
thread 'run_all' panicked at 'assertion failed: errs.len() == n', tests/utils/mod.rs:168:37
```
Looks like you brought down the number of cascading errors in one of the test cases :)

```
$ cargo run tests/runner-tests/decls/test_initializers/8.c
   Compiling saltwater v0.11.0 (/home/joshua/src/rust/saltwater)
    Finished dev [unoptimized + debuginfo] target(s) in 5.74s
     Running `/home/joshua/.local/lib/cargo/target/debug/swcc tests/runner-tests/decls/test_initializers/8.c`
tests/runner-tests/decls/test_initializers/8.c:4:5 warning: implicit int is deprecated and may be removed in a future release
int a[](*fp)() = {0};
    ^^^^^^^^^^
tests/runner-tests/decls/test_initializers/8.c:2:1 error: invalid program: variable has incomplete type 'void'
void i = {1};
^^^^^^
tests/runner-tests/decls/test_initializers/8.c:3:1 error: invalid program: variable has incomplete type 'void'
void i = 1;
^^^^^^
tests/runner-tests/decls/test_initializers/8.c:3:6 error: invalid program: redefinition of 'i'
void i = 1;
     ^
tests/runner-tests/decls/test_initializers/8.c:4:5 error: invalid program: functions cannot return 'int ()'
int a[](*fp)() = {0};
    ^^^^^^^^^^
tests/runner-tests/decls/test_initializers/8.c:4:5 error: invalid program: arrays cannot contain functions (got 'int (int *fp)()'). help: try storing array of pointer to function: (*int (int *fp)()[])(...)
int a[](*fp)() = {0};
    ^^^^^^^^^^
```

src/analyze/expr.rs
```diff
@@ -1313,14 +1313,10 @@ impl Expr {
             // There is probably a better way to do this
             // check not type error and that if rval is a pointer, it is not a type error
```
jyn514, line 1314:
```suggestion
            // don't report cascading errors
```

src/analyze/expr.rs
```diff
@@ -642,6 +666,8 @@ impl PureAnalyzer {
             (Type::Pointer(target, _), _) => ((**target).clone(), left, right),
             // i[p]
             (_, Type::Pointer(target, _)) => ((**target).clone(), right, left),
+            // already type error
+            (Type::Error, _) | (_, Type::Error) => return left,
```
ryankeleti, line 670: Should this split into `return left` and `return right`?

ryankeleti:
Maybe something to look into: `gcc` reports only once for each undeclared identifier in the current function, while `clang` more or less has the behavior from this PR.

src/analyze/expr.rs
jyn514: Yes, that sounds like a good idea.

jyn514:
> Maybe something to look into: `gcc` reports only once for each undeclared identifier in the current function, while `clang` more or less has the behavior from this PR. Also sorry for merge commit!

Ooh, that does sound nice. Maybe the `PureAnalyzer` could store a list of undeclared identifiers and not emit the error if the id is already in the list?

src/analyze/expr.rs
ryankeleti: 6f2e58d93ef6856d887903c62b38d03e30c89e51

ryankeleti:
> Ooh, that does sound nice. Maybe the `PureAnalyzer` could store a list of undeclared identifiers and not emit the error if the id is already in the list? (not necessarily in this PR though)

I like that!

PR #510, opened 2020-08-16
Close #506: Implement preprocessor concatenation

### LINKED ISSUE #506: Can no longer run hello world :(
### Expected behavior

`cargo run tests/runner-tests/hello_world.c` works on linux platforms.

### Code

```c
#include<stdio.h>

int main() {
    puts("Hello, world!");
}
bits/libc-header-start.h:56:4 error: invalid macro: trailing tokens in `#if` expression
#if __GLIBC_USE (IEC_60559_BFP_EXT) || __GLIBC_USE (ISOC2X)
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

hdamron17:
Will close #506. So far I have collected the tokens before and after `##` and passed them to a `concat` function. Now I just need to figure out how to concatenate them.


hdamron17:
Also, I'm probably not working with `replacements` and `pending` properly in [`replace()`](https://github.com/jyn514/saltwater/blob/10019b/src/lex/replace.rs#L157).

hdamron17:
I'm assuming you'll want more tests...

src/lex/cpp.rs
```diff
@@ -1941,4 +1941,50 @@ h",
...
+        assert_concat("5", "e5", Some(Literal(Float("5e5".into()))));
+        assert_concat("1234", ".5", Some(Literal(Float("1234.5".into()))));
+        assert_concat("42", "000", Some(Literal(Int("42000".into()))));
+
+        assert_concat("+", "/", None);
+        assert_concat(r#"'x'"#, r#"'y'"#, None);
+        assert_concat(r#""x""#, r#""y""#, None);
+        assert_concat("0b1", "6", None);
+        assert_concat("/", "/", None); // Not a comment
+    }
```
jyn514, line 1977:
```suggestion
    }
    
    #[test]
    fn gnu_weirdness() {
        // #506
        let program = "
#define __GLIBC_USE(F)	__GLIBC_USE_ ## F
# define __GLIBC_USE_IEC_60559_TYPES_EXT 1 
# define __GLIBC_USE_ISOC2X 1
#if __GLIBC_USE (IEC_60559_BFP_EXT) || __GLIBC_USE (ISOC2X)
	int main() {}
#endif
        ";
        assert_same(program, "int main() {}");
    }
```

src/data/lex.rs
```diff
@@ -231,8 +231,9 @@ pub enum Token {
 
     // Misc
     Ellipsis,
-    StructDeref, // ->
-    Hash,        // #, used for preprocessing
+    StructDeref,    // ->
+    Hash,           // #, used for preprocessing
+    HashHash(bool), // ##, used for preprocessing (the bool is true unless it is created by `# ## #`)
```
jyn514, line 236: It's not clear what the bool is where you use it in the code, might be better to say explicitly (the comment is fine though).
```suggestion
    HashHash { operator: bool }, // ##, used for preprocessing (the bool is true unless it is created by `# ## #`)
```

jyn514, line 236: nit: I'd mildly prefer to either have inconsistent alignment or to put the comment on the line above. Otherwise it's more work next time you need to update the comments.

src/lex/replace.rs
```diff
@@ -166,56 +166,104 @@ pub fn replace(
...
-                        continue;
+        match token {
+            Ok(Locatable {
+                data: ref succeeding_tok,
+                ..
+            }) if pending_hashhash.is_some() => {
+                if matches!(succeeding_tok, Token::Whitespace(_)) {
+                    continue;
+                }
+                let pending_hashhash = pending_hashhash.take().unwrap(); // We just checked that it's some
```
jyn514, line 185: Better to write this as
```rust
match (token, pending_hashhash) {
    (Ok(..), Some(hashhash)) => {
```
That way you don't have an extra `unwrap`.

src/lex/replace.rs
```diff
@@ -166,56 +166,104 @@ pub fn replace(
...
+                let preceding_tok = loop {
+                    match replacements.pop() {
+                        Some(Ok(Locatable {
+                            data: Token::Whitespace(_),
+                            ..
+                        })) => continue,
+                        Some(Ok(Locatable { data: token, .. })) => break token,
+                        None | Some(Err(_)) => {
+                            return wrap_error(&location, CppError::HashHashMissingParameter(true))
+                        }
```
jyn514, line 260: Hmm, in the `Some(Err)` case I think we should return the original error instead of swallowing it.

src/lex/replace.rs
```diff
@@ -439,3 +488,21 @@ fn stringify(args: Vec<Token>) -> Token {
         ret.trim()
     ))]))
 }
+
+fn concat(x: &Token, y: &Token, location: &Location) -> Option<Locatable<Token>> {
+    let mut lexer = Lexer::new(location.file, format!("{}{}", x, y), false);
```
jyn514, line 493: You said doing this means we can get rid of the LiteralParser trait, right? Do you want to do that here or in a follow-up PR?

jyn514 (commented):
So, this looks good to me, I don't see anything wrong. However I'm having doubts about how maintainable this all is ... in particular I don't know how this handles [empty preprocessor arguments](http://port70.net/~nsz/c/c11/n1570.html#6.10.3.3):

> If an argument consists of no preprocessing tokens, the parameter is replaced by a placemarker preprocessing token instead.

I think currently this will give an error, which is not correct. But fixing it sounds really painful and will need lots of rearchitecting.

An idea I came up with was to reuse some of the ideas from https://github.com/eddyb/cprep/. The benefits there are:

- The algorithm is _much_ more clear, the one in saltwater is a bit of a mess
- It runs _before_ the lexer, which I've wanted to do for a while
- It's already written :P

What do you think about merging this as-is (with nits fixed), but eventually transitioning to using `cprep` instead?

cc @eddyb

PR #511, opened 2020-08-17
Replce shared_str with arcstr

### LINKED ISSUE #509: Switch from shared_str to arcstr
`arcstr` is better maintained and better tested (especially important because both `shared_str` and `arcstr` have unsafe code).

Relevant code: https://github.com/jyn514/saltwater/blob/master/Cargo.toml#L43, https://github.com/jyn514/saltwater/blob/master/src/lex/mod.rs#L121, but you should be able to find most of it from the compile errors.

cc @thomcc

Stupremee:
Resolves #509

### DIFF
--- a/src/lex/mod.rs
+++ b/src/lex/mod.rs
@@ -1,5 +1,4 @@
 use std::convert::{TryFrom, TryInto};
-use std::rc::Rc;
 
 use codespan::FileId;
 
@@ -36,7 +35,7 @@ type LexResult<T = Token> = Result<T, Locatable<LexError>>;
 #[derive(Debug)]
 pub struct Lexer {
     location: SingleLocation,
-    chars: Rc<str>,
+    chars: ArcStr,
     /// used for 2-character tokens
     current: Option<char>,
     /// used for 3-character tokens
@@ -78,7 +77,7 @@ pub(crate) struct SingleLocation {
 
 impl Lexer {
     /// Creates a Lexer from a filename and the contents of a file
-    pub fn new<S: Into<Rc<str>>>(file: FileId, chars: S, debug: bool) -> Lexer {
+    pub fn new<S: Into<ArcStr>>(file: FileId, chars: S, debug: bool) -> Lexer {
         Lexer {
             given_newline_error: false,
             debug,
@@ -122,7 +121,7 @@ impl Lexer {
         use std::ops::Range;
 
         let range: Range<usize> = self.span(span_start).span.into();
-        ArcStr::from(self.chars.clone()).substr(range)
+        self.chars.substr(range)
     }
 
     /// Parse a number literal, given the starting character and whether floats are allowed.
--- a/src/lex/mod.rs
+++ b/src/lex/mod.rs
@@ -9,7 +9,7 @@ use super::data::{
     *,
 };
 use super::intern::InternedStr;
-use arcstr::ArcStr;
+use arcstr::{ArcStr, Substr};
 
 mod cpp;
 mod files;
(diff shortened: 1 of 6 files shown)


src/lex/mod.rs
```diff
@@ -118,14 +118,14 @@ impl Lexer {
...
-                s.get::<Range<usize>>(self.span(span_start).span.into())
-                    .unwrap_or("")
-            })
-            .unwrap()
+
+        let chars = self.chars.clone();
+        let chars = chars
+            .get::<Range<usize>>(self.span(span_start).span.into())
+            .unwrap_or("");
+        ArcStr::from(chars)
```
jyn514, line 128: Doesn't this copy the entire token? The whole point of using Rc is so that we don't burn up memory by copying each token many times.

hdamron17: This would copy the string. You want to clone `self.chars` since it's Rc then substring on that using ArcStr's substring function.

Stupremee: Which `substring` function?

hdamron17: Oh maybe the sub string was an open PR. You could wait for it to go through or what you can do is if you clone the Rc into an RcStr then do a `get` or something similar which gives you a `&str` and you can rebind that with the ArcStr or something like that. I can give you better details when I'm back at my computer.

hdamron17: This is the PR- it's merged but not in a release yet. https://github.com/thomcc/arcstr/pull/12

hdamron17: So this means you'll also want to use `Substr` type instead of a raw `ArcStr`.

jyn514: To unblock this, can you use the git repo for now? That way we can see if this at least works before having to wait for the release.

Stupremee: Oh god. that will take time to refactor everything, because now almost everything has to use `Substr`

src/lex/mod.rs
```diff
@@ -118,14 +118,11 @@ impl Lexer {
...
+    fn slice(&self, span_start: u32) -> Substr {
         use std::ops::Range;
 
-        let chars = self.chars.clone();
-        let chars = chars
-            .get::<Range<usize>>(self.span(span_start).span.into())
-            .unwrap_or("");
-        ArcStr::from(chars)
+        let range: Range<usize> = self.span(span_start).span.into();
+        ArcStr::from(self.chars.clone()).substr(range)
```
jyn514, line 125: This is still missing the point. ArcStr::from copies the input. You should make `self.chars` an ArcStr itself so that you only call `from` once in the whole program.

jyn514: Hmm... Actually I think it might not copy if you're creating it from an Rc. Can you use `Rc::clone` here to make sure that's the generic impl that's chosen?

Stupremee: Nah, that's actually a clone of the chars.

hdamron17:
To answer the previous question about `literal!`, [here is the explanation](https://github.com/thomcc/arcstr/blob/d7a8632/src/arc_str.rs#L56). To paraphrase, it means anything created as a literal is static and `.clone()` does effectively nothing.

hdamron17:
Also, possibly you can add a test that ensures the pointer returned by `Lexer::slice` is identical to the lexer's `ArcStr`. This would use `pointer_eq!` provided by `arcstr`, I believe.

src/lex/mod.rs
thomcc: You were correct here, that it will copy even if creating from an Rc. We don't reuse the stdlib Arc/Rc impl for a lot of reasons: FFI safety, avoiding needing to worry about mutability or Weak, etc, ensuring our ArcStr is both a single pointer without adding an extra indirection, transparent/cheap support for literals, ... Honestly probably more.

src/lex/mod.rs
thomcc: Sorry. I think I've done this refactor for y'all though, it made me notice a couple missing bits in the API too.

Stupremee:
Closed in favor of #514

PR #512, opened 2020-08-17
Don't emit error in binary promote if left type error

ryankeleti:
Oops, found another! `x &= 1`, `x |= 1`, etc. will report two errors instead of one.


jyn514:
Can you add a test for this?

ryankeleti:
On it, I'll add some for the others too

tests/runner-tests/expr/undecl/0.c
```diff
@@ -0,0 +1,11 @@
+// errors: 7
```
jyn514, line 1: Since 'undeclared' is only used in one or two places in the compiler, can you change the directory name to use the full word instead of abbreviating it? `expr` is abbreviated because it's used so much but I had to think about `undecl` for a second.

ryankeleti: a0ba82511af385495f766899764dac8c7913db97

PR #514, opened 2020-08-19
Convert code from shared_string to arcstr

thomcc:
You asked me to in https://github.com/thomcc/arcstr/issues/13#issuecomment-674550811 and it's a good excuse to test out the API of ArcStr and make sure there are no gaps.

Note that arcstr's types show up in the public API as `Into<ArcStr>`, whereas shared_string did not, since we don't use `Arc<str>`/`Rc<str>` under the hood. Without without this you'd essentially have no reason to use `arcstr` as each substr would be independent (actually, there are probably ways around this if you really need it but it would require additional string copies).

Also note that your benchmarks don't compile for me with or without these changes. I'll see if the fix is trivial but if not will leave it to y'all.

Also the tests fail for me with and without these changes because someone somewhere can't find `puts`. Maybe my `stdio.h` is weird or something idk.

### DIFF
--- a/src/data/lex.rs
+++ b/src/data/lex.rs
@@ -7,7 +7,7 @@ use proptest_derive::Arbitrary;
 use crate::data::hir::BinaryOp;
 use crate::intern::InternedStr;
 
-use shared_str::RcStr;
+use arcstr::Substr;
 
 // holds where a piece of code came from
 // should almost always be immutable
@@ -166,11 +166,11 @@ pub enum ComparisonToken {
 #[derive(Clone, Debug)]
 pub enum LiteralToken {
     // literals
-    Int(RcStr),
-    UnsignedInt(RcStr),
-    Float(RcStr),
-    Str(Vec<RcStr>),
-    Char(RcStr),
+    Int(Substr),
+    UnsignedInt(Substr),
+    Float(Substr),
+    Str(Vec<Substr>),
+    Char(Substr),
 }
 
 impl PartialEq for LiteralToken {
@@ -404,15 +404,15 @@ impl std::fmt::Display for LiteralToken {
             Int(i) => write!(f, "{}", i),
             UnsignedInt(u) => write!(f, "{}", u),
             Float(n) => write!(f, "{}", n),
-            Str(rcstr) => {
-                let joined = rcstr
-                    .iter()
-                    .map(RcStr::as_str)
-                    .collect::<Vec<_>>()
+            Str(s) => {
+                let joined = s
+                    // .iter()
+                    // .map(Substr::as_str)
+                    // .collect::<Vec<_>>()
                     .join(" ");
                 write!(f, "{}", joined)
             }
-            Char(rcstr) => write!(f, "{}", rcstr.as_str()),
+            Char(s) => write!(f, "{}", s),
         }
     }
 }
@@ -463,8 +463,8 @@ impl From<ComparisonToken> for Token {
 #[cfg(test)]
 mod proptest_impl {
     use super::LiteralToken;
+    use arcstr::Substr;
     use proptest::prelude::*;
-    use shared_str::RcStr;
 
     impl Arbitrary for LiteralToken {
         type Parameters = ();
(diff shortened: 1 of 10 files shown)


Stupremee:
Yes, the `hello_world.c` test only really works in the CI.

hdamron17:
This is already in progress in #511, but I'm interested to see how the two implementations compare.

Also, the issue you had with the tests is (I'm guessing) `hello_world.c` which has `#include <stdio.h>`. This works on Travis, but it depends on your local standard library which may have features not yet supported by saltwater. You can just remove `#include <stdio.h>` and declare `int puts(const char *);`. Hopefully soon we'll be able to compile `stdio.h` on any machine.

Stupremee:
I think we should merge this PR instead of #511 , because it already uses `arcstr::format` and also has ported the benchmarks

src/data/lex.rs
```diff
@@ -404,15 +404,15 @@ impl std::fmt::Display for LiteralToken {
...
-                    .map(RcStr::as_str)
-                    .collect::<Vec<_>>()
+            Str(s) => {
+                let joined = s
+                    // .iter()
+                    // .map(Substr::as_str)
+                    // .collect::<Vec<_>>()
                     .join(" ");
                 write!(f, "{}", joined)
             }
```
jyn514, line 414:
```suggestion
            Str(s) => write!(f, "{}", s.join(" ")),
```

thomcc:
I can't make the `parens` bench compile for reasons unrelated to my changes, and while I can make `examples` work the changes are trivial, unrelated to these changes, and seem better handled in a dedicated "fix the benches" patch.

That said, if you want the diff for benches/examples.rs it's just:

```diff
diff --git a/benches/examples.rs b/benches/examples.rs
index 1f89a46..7eb62d2 100644
--- a/benches/examples.rs
+++ b/benches/examples.rs
@@ -1,11 +1,11 @@
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
-use rcc::{compile, initialize_jit_module};
+use saltwater::{compile, initialize_jit_module};
 
 const FACTORIAL: &str = include_str!("../tests/runner-tests/factorial.c");
 const FIBONACCI: &str = include_str!("../tests/runner-tests/fibonacci.c");
 
 fn examples(c: &mut Criterion) {
-    let opts = rcc::Opt {
+    let opts = saltwater::Opt {
         filename: "<benchmark>".into(),
         ..Default::default()
     };
@@ -13,7 +13,7 @@ fn examples(c: &mut Criterion) {
     group.bench_function("rcc", |b| {
         b.iter(|| {
             let module = initialize_jit_module();
-            black_box(compile(module, FIBONACCI, &opts))
+            black_box(compile(module, FIBONACCI, opts.clone()))
         });
     });
     group.finish();
@@ -21,7 +21,7 @@ fn examples(c: &mut Criterion) {
     group.bench_function("rcc", |b| {
         b.iter(|| {
             let module = initialize_jit_module();
-            black_box(compile(module, FACTORIAL, &opts))
+            black_box(compile(module, FACTORIAL, opts.clone()))
         });
     });
     group.finish();
```

I do think it would be nice if `examples.rs` could use `arcstr::literal!(include_str!(...))` for FACTORIAL and FIBONACCI, and pass those into `compile` but that requires changing the signature of even more stuff (e.g. `fn compile`) and as is, this probably a more realistic test of functionality. It also seems unlikely to matter much for performance.

src/data/lex.rs
thomcc: Thanks, forgot to clean this up and wasnt sure it would compile.

thomcc:
> Users will be able to pass in normal Strings wherever the code says impl From<T: Into<ArcStr>>?

Yes, they should be able to pass `String`, `&str`, `Cow<'a, str>`, `Rc<str>`, or really anything that seemed sufficiently string-like... If there's a missing From `arcstr` needs file a bug.

> Can you add the test mentioned in #511 (comment)?

Yes but I think I need a little more guidance, sorry.

jyn514:
Yeah definitely don't worry about the benches, I need to add CI for them so they stop breaking.

I think for ArcStr the test would look something like this:
```
let src = ArcStr::from("int main() {}");
let mut files = Files::new();
let file_id = files.add("<bench>", ArcStr::clone(&src));
let mut lexer = Lexer::new(file_id, ArcStr::clone(&str), false);
let token = lexer.next().unwrap().unwrap().data;
assert!(token.ptr_eq(str));
```

thomcc:
> I think for ArcStr the test would look something like this:

Already figured it out, although mine is slightly different, let me know if I should change it.

thomcc:
> If there's a missing From arcstr needs file a bug.

Realistically the only case I can think of where it could be an issue is if they have an `ArcStr` from a different version due to semver compat. We won't accept that. That said they're unlikely to have a good time in a case like this no matter what, and can work around it with a `.as_str()` or something. I also don't intend to do breaking changes all the time or anything.

jyn514:
Yeah if you have two different versions of ArcStr something's already fishy and either I or the downstream user should update. No need to worry about that case.

jyn514:
Thanks for the PR!

PR #515, opened 2020-09-08
Implement proper REPL

### LINKED ISSUE #372: Use rustyline for REPL

Stupremee:
It's the last pr for a repl, I swear :sweat_smile: 

Resolves #372


saltwater-repl/src/commands.rs
```diff
@@ -0,0 +1,27 @@
+use crate::repl::Repl;
+use std::collections::HashMap;
+
+pub fn default_commands() -> HashMap<&'static str, fn(&mut Repl, &str)> {
+    let mut map = HashMap::<&'static str, fn(&mut Repl, &str)>::new();
+    map.insert("help", help_command);
+    map.insert("h", help_command);
+    map.insert("quit", quit_command);
+    map.insert("q", quit_command);
+    map
```
jyn514, line 10: Hmm, I was going to suggest using `map!` but it's not public and I don't know if I want to make it public ... probably not worth it for a map this small.

saltwater-repl/src/commands.rs
```diff
@@ -0,0 +1,27 @@
...
+
+fn help_command(_repl: &mut Repl, _args: &str) {
+    print!(
+        "\
+Available commands:
+    {p}help|h    Shows this message
+    {p}quit|q    Quits the repl
+",
+        p = crate::repl::PREFIX
+    );
```
jyn514, line 21: Does rustyline have a way to do this automatically? I want to avoid having to keep this in sync with `default_commands`.

saltwater-repl/src/helper.rs
```diff
@@ -0,0 +1,126 @@
...
+                },
+                _ => continue,
+            }
+        }
+
+        if stack.is_empty() {
+            Ok(ValidationResult::Valid(None))
+        } else {
+            Ok(ValidationResult::Incomplete)
+        }
```
jyn514, line 67: Rather than trying to guesstimate this, wouldn't it be better to use `saltwater::check_semantics(input).is_ok()`? Maybe even say it's valid for anything that's not a syntax error so that semantic errors show the full message.

saltwater-repl/src/main.rs
```diff
@@ -0,0 +1,17 @@
+//! The repl implementation for saltwater.
+#![deny(rust_2018_idioms)]
```
jyn514, line 2: This is such a badly named lint 😆

jyn514 (commented):
Hmm, the other workspace members don't use a nested `src` directory. I'm not _super_ tied to that, but is there a reason you added it here? The only thing at the top-level looks like Cargo.toml

Stupremee:
> Hmm, the other workspace members don't use a nested `src` directory. I'm not _super_ tied to that, but is there a reason you added it here? The only thing at the top-level looks like Cargo.toml

Ah I see, that's just how `cargo new` generated it. I will change it :+1:

saltwater-repl/src/commands.rs
Stupremee: No, it doesn't event have support for commands, but I can write some code to generate it.

saltwater-repl/src/helper.rs
Stupremee: That sounds like a good idea

saltwater-repl/src/commands.rs
Stupremee: This map will grow over time (I guess), so probably it is worth it

saltwater-repl/src/main.rs
Stupremee: Should I remove it?

jyn514: No no, you're fine, just that `deny(idioms)` looks weird.

Stupremee: Yea that's true

saltwater-repl/src/commands.rs
Stupremee: Okay, this is unessecarry because there's now a `cmds` macro

jyn514:
This breaks when compiling without `--all-features`, although I'm not sure why.
```
error[E0432]: unresolved import `saltwater_codegen::compile_decl`
 --> saltwater-repl/repl.rs:7:25
  |
7 | use saltwater_codegen::{compile_decl, initialize_jit_module, JIT};
  |                         ^^^^^^^^^^^^ no `compile_decl` in the root
error: aborting due to previous error
For more information about this error, try `rustc --explain E0432`.
error: could not compile `saltwater-repl`.
To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error[E0063]: missing field `jit` in initializer of `saltwater_parser::Opt`
   --> src/main.rs:361:14
    |
361 |         opt: Opt {
    |              ^^^ missing `jit`
error: aborting due to previous error
```

Stupremee:
Yea the first error is fixed by my latest commit, but the other error is weird

jyn514:
```
error[E0063]: missing field `jit` in initializer of `saltwater_parser::Opt`
   --> src/main.rs:361:14
    |
361 |         opt: Opt {
    |              ^^^ missing `jit`
```
You're passing `--jit` to the codegen crate, but not to the saltwater binary. I think this only popped up now because previously `clippy` was ignoring the `jit` feature completely for codegen. I'll fix this, one sec.

saltwater-repl/commands.rs
```diff
@@ -0,0 +1,54 @@
...
+                cmd.description
+            )?;
+        }
+        Ok::<String, Box<dyn std::error::Error>>(buf)
+    };
+    inner().expect("failed to generate help message")
+}
+
+fn help_command(repl: &mut Repl, _args: &str) {
+    print!("{}", repl.help_message);
```
jyn514, line 48: This needs to flush or it won't show up:
```suggestion
    print!("{}", repl.help_message);
    std::io::stdout().flush();
```
Is there a reason you're not using `println!`?

saltwater-repl/main.rs
```diff
@@ -0,0 +1,17 @@
...
+mod commands;
+mod helper;
+mod repl;
+
+fn main() {
+    let mut repl = repl::Repl::new();
+    match repl.run() {
+        Ok(_) => {}
+        Err(err) => {
+            println!("error: {}", err);
```
jyn514, line 13: Hmm, these errors are only sometimes showing up.
```
>> "a";
error: expression returns unsupported type: Array(Char(true), Fixed(2))
>> @
# I pressed ctrl+c
>> x = 1

>>
```

saltwater-repl/repl.rs
```diff
@@ -0,0 +1,217 @@
...
+        let expr = analyze_expr(code)?;
+        let expr_ty = expr.ctype.clone();
+        let decl = wrap_expr(expr);
+        let module = compile_decl(module, decl, Default::default())?;
+
+        let mut jit = JIT::from(module);
+        jit.finalize();
+        let fun = jit
+            .get_compiled_function("execute")
+            .expect("this is not good.");
```
jyn514, line 141:
```suggestion
            .expect("wrap_expr should create a function named `execute`");
```

saltwater-parser/analyze/mod.rs
```diff
@@ -139,6 +139,15 @@ impl PureAnalyzer {
     pub fn warnings(&mut self) -> VecDeque<CompileWarning> {
         std::mem::take(&mut self.error_handler.warnings)
     }
+
+    /// Return all errors seen so far.
+    ///
+    /// These errors are consumed and will not be returned if you call
+    /// `errors()` again.
+    pub fn errors(&mut self) -> VecDeque<Locatable<Error>> {
```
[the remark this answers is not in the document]
Stupremee: An `Errors` traits sound good. It can also provide methods to take warnings, or just access them without removing them.

saltwater-repl/commands.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub names: &'static [&'static str],
+    pub description: &'static str,
+    pub action: fn(&mut Repl, &str),
+}
+
+pub fn default_commands() -> Vec<Command> {
+    cmds! {
+        "help", "h"; "Shows this help message" => help_command,
+        "quit", "q"; "Quits the repl" => quit_command,
+    }
```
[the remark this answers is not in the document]
Stupremee: Nope that looks fine.

jyn514:
Opened https://github.com/bytecodealliance/wasmtime/pull/2199

saltwater-repl/helper.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl Highlighter for ReplHelper {
+    fn highlight<'l>(&self, line: &'l str, pos: usize) -> Cow<'l, str> {
+        // TODO: Syntax highlighting.
+        self.highlighter.highlight(line, pos)
+    }
+
+    fn highlight_hint<'h>(&self, hint: &'h str) -> Cow<'h, str> {
+        Cow::Owned(hint.dimmed().to_string())
```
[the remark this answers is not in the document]
Stupremee: Yea it is. It would save so many allocations

saltwater-repl/commands.rs
```diff
@@ -0,0 +1,54 @@
...
+        Ok::<String, Box<dyn std::error::Error>>(buf)
+    };
+    inner().expect("failed to generate help message")
+}
+
+fn help_command(repl: &mut Repl, _args: &str) {
+    print!("{}", repl.help_message);
+}
+
+fn quit_command(repl: &mut Repl, _args: &str) {
```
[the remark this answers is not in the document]
Stupremee: That prevents it to be used as an `action` in a `Command` because `action` has no return type.

saltwater-repl/commands.rs
Stupremee: Nope not really

saltwater-repl/main.rs
Stupremee: These lines only print the error that is returned by `rustyline` if something terminal related fails. 
The input isn't valid, because the `Validator` checks if the input is a valid expression.

saltwater-repl/commands.rs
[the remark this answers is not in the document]
jyn514: Ah well, fair enough.

saltwater-repl/helper.rs
```diff
@@ -42,7 +42,11 @@ impl Highlighter for ReplHelper {
 
 impl Validator for ReplHelper {
     fn validate(&self, ctx: &mut ValidationContext<'_>) -> rustyline::Result<ValidationResult> {
-        let result = crate::repl::analyze_expr(ctx.input());
+        if ctx.input().starts_with(crate::repl::PREFIX) {
+            return Ok(ValidationResult::Valid(None));
```
jyn514, line 46: Shouldn't this also check that the command is recognized?

saltwater-repl/repl.rs
```diff
@@ -165,7 +179,7 @@ impl Repl {
             Type::Void => execute!(fun, ()),
 
             // TODO: Implement execution for more types
-            ty => println!("error: expression returns unsupported type: {:?}", ty),
+            ty => println!("error: expression has an  unsupported type: {:?}", ty),
```
jyn514, line 182:
```suggestion
            ty => println!("error: expression has an unsupported type: {:?}", ty),
```

saltwater-repl/repl.rs
```diff
@@ -0,0 +1,231 @@
...
+/// Takes a function pointer and transmute it into a function
+/// that returns the given type. Then executes it and returns the result.
+///
+/// # Safety
+///
+/// `$fun` has to be a valid pointer to a function that returns the `$ty`.
+macro_rules! execute {
+    ($fun:ident, $ty:tt) => {
+        unsafe {
+            let execute: unsafe extern "C" fn() -> $ty = std::mem::transmute($fun);
```
jyn514, line 28: Let's also specify the source type.
```suggestion
            let execute = std::mem::transmute<*const u8, unsafe extern "C" fn() -> $ty> ($fun);
```

saltwater-repl/main.rs
jyn514: Right, but this isn't reporting the error at _all_, which is a bug. Do you know how to fix that? I'm not sure entirely where the syntax errors are bubbling up - maybe in `execute_code`?

Stupremee: The errors are reported in `process_line`

Stupremee:
Should I wait until `target-lexicon` is updated?

jyn514:
For now let's just add `#[cfg(not(target_pointer_width = "64"))] compile_error!("only x86_64 is supported");`, and then we can make it a run-time error instead once cranelift is updated.

saltwater-repl/repl.rs
```diff
@@ -0,0 +1,226 @@
...
+                    action(self, &trimmed_line[name.len()..])
+                }
+                None => println!("unknown command '{}'", name),
+            }
+        } else {
+            match self.execute_code(line) {
+                Ok(_) => {}
+                Err(err) => {
+                    // TODO: Proper error reporting
+                    println!("error: {}", err.data);
```
jyn514, line 132: This bit doesn't seem to be working (as mentioned above):
```
>> @
;
# no output
```

Stupremee: because it's not a valid expression and thus get declined by rustyline.

PR #516, opened 2020-09-13
Run clippy on non-default features, too

jyn514:
This fixes the CI failure in https://github.com/jyn514/saltwater/pull/515#issuecomment-691684770 and is generally good practice.


jyn514:
This also removes `#include<stdio.h>` from the test suite because it's causing tests to fail locally depending on the machine you run things on.

